PR #11, opened 2015-06-13
[WIP] Add filenames source

fmoralesc:
I'm not sure the behavior of this source is ideal yet, please give me suggestions. 

It probably needs to take into account the context, so completing from

~~~
rplugin/python3/deopl|
                     ^
~~~
gives
~~~
rplugin/python3/deoplete|
                        ^
~~~
What is the proper way to do this?

### DIFF
--- a/rplugin/python3/deoplete/sources/files.py
+++ b/rplugin/python3/deoplete/sources/files.py
@@ -1,0 +1,66 @@
+#=============================================================================
+# FILE: files.py
+# AUTHOR: Felipe Morales <hel.sheep at gmail.com>
+# License: MIT license  {{{
+#     Permission is hereby granted, free of charge, to any person obtaining
+#     a copy of this software and associated documentation files (the
+#     "Software"), to deal in the Software without restriction, including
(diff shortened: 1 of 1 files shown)


Shougo:
Yes, you should parse input text.
Unfortunlatelly, I don't implement complete position feature.
You should wait until it.

fmoralesc:
OK, I'll wait.

Shougo:
I have implemented source complete position feature.
Please implement the source using it.

fmoralesc:
I updated the source to use the complete position feature (great work on that, it's really nice to use :+1:).

fmoralesc:
I was thinking: should this return the paths macthing the prefix, as it does now, or should it return all files in the dirname, so a fuzzy matcher could get better candidates?

Shougo:
I think current behavior is better.

Shougo:
Merged.  Thanks.

If you want to update it, please create another pull request.

Shougo:
I think file name is better.
I will rename files to file.

xHN35RQ:
:+1:

PR #19, opened 2015-06-20
Cache tags

fmoralesc:
Should improve performance of tag source.

### DIFF
--- a/rplugin/python3/deoplete/sources/tag.py
+++ b/rplugin/python3/deoplete/sources/tag.py
@@ -24,7 +24,11 @@
 #=============================================================================
 
 import re
+from os.path import getmtime
+from collections import namedtuple
 from .base import Base
+
+TagsCacheItem = namedtuple('TagsCacheItem', 'mtime tags')
 
 class Source(Base):
     def __init__(self):
@@ -32,14 +36,22 @@ class Source(Base):
 
         self.mark = '[T]'
 
+        self.cache = {}
+
     def gather_candidates(self, vim, context):
         candidates = []
-
         for tags_file in vim.eval(
                 'map(tagfiles(), "fnamemodify(v:val, \\":p\\")")'):
-            with open(tags_file, 'r', errors='replace') as f:
-                candidates += parse_tags(f)
-        return [{ 'word': x } for x in list(set(candidates))]
+            mtime = getmtime(tags_file)
+            if tags_file not in self.cache or \
+                    self.cache[tags_file].mtime != mtime:
+                with open(tags_file, 'r', errors='replace') as f:
+                    new_candidates = parse_tags(f)
+                    candidates += new_candidates
+                self.cache[tags_file] = TagsCacheItem(mtime, new_candidates)
+            else:
+                candidates += self.cache[tags_file].tags
+        return [{ 'word': x } for x in candidates]
 
 def debug(vim, msg):
     vim.command('echomsg string("' + str(msg) + '")')
@@ -50,5 +62,5 @@ def parse_tags(f):
 
     for l in f.readlines():
         candidates += p.findall(l)
-    return candidates
+    return list(set(candidates))
 


rplugin/python3/deoplete/sources/tag.py
```diff
@@ -50,5 +61,5 @@ def parse_tags(f):
 
     for l in f.readlines():
         candidates += p.findall(l)
-    return candidates
+    return list(set(candidates))
```
ZyX-I: I am wondering what’s the point of list(set())? If you need to randomize list of candidates there is `random.shuffle` which makes it *explicit* what you are doing and does not put unneded requirements on the elements of the list.

fmoralesc: No, the point is to avoid duplicates.

fmoralesc: Of course, it might not be the best way to do it ;)

ZyX-I:
Your PR adds cache, how did code that avoids candidates appear in it? If it was my project I would say that you *must* do this at least in a separate commit entirely (much better: in a separate PR).

fmoralesc:
The code that avoids candidates is just a cleanup of the previous behavior (the tag source always prevented duplicates); I thought you were suggesting that previously. I can split that from this PR, I just didn't want to clutter things up. I'll wait for @Shougo's comments.

rplugin/python3/deoplete/sources/tag.py
```diff
@@ -24,31 +24,46 @@
...
+            else:
+                candidates += self.cache[tags_file].tags
+        return [{ 'word': x } for x in candidates]
 
 def debug(vim, msg):
     vim.command('echomsg string("' + str(msg) + '")')
 
 def parse_tags(f):
-    p = re.compile('^[a-zA-Z_]\w*(?=\t)')
+    p = re.compile('[a-zA-Z_]\w*(?=\t)')
```
ZyX-I: Why? This variant is going to catch more things as tags. In some cases it will *definitely* catch non-tags or partial tags. And this change also

1. Has absolutely nothing to do with adding caching (commit 1, PR purpose).
2. Has absolutely nothing to do with changing method of avoiding candidates (commit 2 in which it is present).

ZyX-I:
> I thought you were suggesting that previously.

I did not suggest that, I was wondering what one specific change was doing.

rplugin/python3/deoplete/sources/tag.py
fmoralesc: > Why? This variant is going to catch more things as tags.

No, it won't. Notice I'm using `re.match()` below, the `^` is implicit in that.

ZyX-I: Understood. This is definitely cleaner, though I keep always forgetting about the fact (in most languages including VimL and lua “match” means “match anywhere”).

fmoralesc: Yeah, I always forget about that myself.

rplugin/python3/deoplete/sources/tag.py
```diff
@@ -24,31 +24,46 @@
 #=============================================================================
 
 import re
+from os.path import getmtime
+from collections import namedtuple
 from .base import Base
 
+TagsCacheItem = namedtuple("TagsCacheItem", "mtime tags")
```
Shougo: Please use single quotes.

rplugin/python3/deoplete/sources/tag.py
```diff
@@ -24,31 +24,46 @@
...
                 'map(tagfiles(), "fnamemodify(v:val, \\":p\\")")'):
-            with open(tags_file, 'r', errors='replace') as f:
-                candidates += parse_tags(f)
-        return [{ 'word': x } for x in list(set(candidates))]
+            mtime = getmtime(tags_file)
+            if tags_file not in self.cache or \
+                    self.cache[tags_file].mtime != mtime:
+                with open(tags_file, 'r', errors='replace') as f:
+                    candidates += parse_tags(f)
+                self.cache[tags_file] = TagsCacheItem(mtime, candidates)
```
Shougo: This seems bug.
You should save `parse_tags(f)` resulted value and set as `TagsCacheItem`.

rplugin/python3/deoplete/sources/tag.py
```diff
@@ -24,31 +24,46 @@
...
     candidates = []
 
-    for l in f.readlines():
-        candidates += p.findall(l)
+    for line in f.readlines():
+        match = p.match(line)
+        if match:
+            tag = match.group()
+            if tag not in candidates:
+                candidates += tag
```
Shougo: It is for `uniq()`?
I think to use `Set` is better.

Shougo:
Thanks.  I reviewed it.

rplugin/python3/deoplete/sources/tag.py
fmoralesc: You are totally right, sorry!

rplugin/python3/deoplete/sources/tag.py
fmoralesc: Ok, I'll revert to the original version :+1:

fmoralesc:
Fixed the issues and reverted the second commit. Thanks for taking a look, that was an ugly brain fart.

Shougo:
Thanks.  LGTM.

PR #21, opened 2015-06-23
fix start_length = 1

HerringtonDarkholme:
as reported in https://github.com/Shougo/deoplete.nvim/issues/20


Shougo:
Thanks.  Fixed.

PR #22, opened 2015-06-23
sources/file: new method to get completion position

fmoralesc:
Use the longest path that can be constructed from context['input']

This doesn't need to parse '&isfname', is simpler, and allows completion
of filenames containing spaces.

Fixes issues described here :
https://github.com/Shougo/deoplete.nvim/commit/a65c1703c4fabf1dc159f30ca64e259f6fa6b489#commitcomment-11823924

### DIFF
--- a/rplugin/python3/deoplete/sources/file.py
+++ b/rplugin/python3/deoplete/sources/file.py
@@ -26,18 +26,17 @@
 
 import re
 import os
+from os.path import exists, dirname
 from glob import glob
 from .base import Base
 
-def parse_isfname(isfname):
-    values = []
-    parts = isfname.split(',')
-    for part in parts:
-        if re.match('\d+-\d+', part):
-            '-'.join(map(lambda x: chr(int(x)), part.split('-')))
-        else:
-            values += part
-    return ''.join(values)
+def longest_path_that_exists(input_str):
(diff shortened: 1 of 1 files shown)


Shougo:
OK.  It seems better.
But can you change the complete position like neocomplete file source?
Please test neocomplete.
Current file source complete position behavior is different.

kopischke:
Looks like the much saner path to tread to me – `isf` is something you only want to touch with a long, fireproof stick (preferably one remote manipulated while you hide in a fallout shelter) :).

fmoralesc:
I updated this with some fixes. I'm not sure what in neocomplete's behavior is different, and why it is preferable, I'd be glad if @Shougo could point that out to me.

Shougo:
Merged.

OK, I explain it.

If `/foo/bar/|` : (`|` is cursor position)

neocomplete completes `/foo/bar/baz` from last slash.
current deoplete file source completes `/foo/bar/baz` from first slash.

I think neocomplete behavior is better.
Because, neocomplete file source also can complete from other source(For example: buffer).
And complete string is not long.

Shougo:
Another improvement:

The file source should convert `\` to `/`.
This routine seems not supported Windows path separator (backslash).

PR #25, opened 2015-07-16
Update for breaking change in python-client 0.0.37

bfredl:
Remove bytes to str conversion, as this is done automatically now. Ref neovim/python-client#129


Shougo:
Thanks.

PR #37, opened 2015-08-28
Add a default omni completion pattern for Go

svanharmelen:
Adding this will make using deoplete with Go a little easier to setup :smiley:


Shougo:
OK.  Thanks.

PR #79, opened 2015-11-19
Restore user's completeopt setting

blueyed:
The buffer's completeopt setting should get restored after deoplete is
done.

This is required when using deoplete on demand.

### DIFF
--- a/autoload/deoplete/handlers.vim
+++ b/autoload/deoplete/handlers.vim
@@ -72,15 +72,14 @@ function! s:completion_begin(event) abort "{{{
 endfunction"}}}
 
 function! s:on_insert_leave() abort "{{{
+  if exists('g:deoplete#_context.saved_completeopt')
+    let &completeopt = g:deoplete#_context.saved_completeopt
+  endif
   let g:deoplete#_context = {}
 endfunction"}}}
 
 function! s:complete_done() abort "{{{
   let g:deoplete#_context.position = getpos('.')
-  if exists('b:deoplete_saved_completeopt')
-    let &completeopt = b:deoplete_saved_completeopt
-    unlet b:deoplete_saved_completeopt
-  endif
 endfunction"}}}
 
 " vim: foldmethod=marker
--- a/autoload/deoplete/handlers.vim
+++ b/autoload/deoplete/handlers.vim
@@ -77,6 +77,10 @@ endfunction"}}}
 
 function! s:complete_done() abort "{{{
   let g:deoplete#_context.position = getpos('.')
+  if exists('b:deoplete_saved_completeopt')
+    let &completeopt = b:deoplete_saved_completeopt
+    unlet b:deoplete_saved_completeopt
+  endif
 endfunction"}}}
 
 " vim: foldmethod=marker
--- a/autoload/deoplete/handlers.vim
+++ b/autoload/deoplete/handlers.vim
@@ -51,8 +51,6 @@ function! s:completion_begin(event) abort "{{{
   " Save the previous position
   let g:deoplete#_context.position = context.position
 
-  call deoplete#mappings#_set_completeopt()
-
   " Call omni completion
   for filetype in context.filetypes
     for pattern in deoplete#util#convert2list(
@@ -63,6 +61,7 @@ function! s:completion_begin(event) abort "{{{
       if deoplete#util#is_eskk_convertion()
             \ || (pattern != '' && &l:omnifunc != ''
             \ && context.input =~# '\%('.pattern.'\)$')
+        call deoplete#mappings#_set_completeopt()
         call feedkeys("\<C-x>\<C-o>", 'n')
         return
       endif
--- a/autoload/deoplete/handlers.vim
+++ b/autoload/deoplete/handlers.vim
@@ -74,6 +74,7 @@ endfunction"}}}
 function! s:on_insert_leave() abort "{{{
   if exists('g:deoplete#_context.saved_completeopt')
     let &completeopt = g:deoplete#_context.saved_completeopt
+    unlet g:deoplete#_context.saved_completeopt
   endif
   let g:deoplete#_context = {}
 endfunction"}}}
--- a/autoload/deoplete/handlers.vim
+++ b/autoload/deoplete/handlers.vim
@@ -51,6 +51,8 @@ function! s:completion_begin(event) abort "{{{
   " Save the previous position
   let g:deoplete#_context.position = context.position
 
+  call deoplete#mappings#_set_completeopt()
+
   " Call omni completion
   for filetype in context.filetypes
     for pattern in deoplete#util#convert2list(
@@ -61,7 +63,6 @@ function! s:completion_begin(event) abort "{{{
       if deoplete#util#is_eskk_convertion()
             \ || (pattern != '' && &l:omnifunc != ''
             \ && context.input =~# '\%('.pattern.'\)$')
-        call deoplete#mappings#_set_completeopt()
         call feedkeys("\<C-x>\<C-o>", 'n')
         return
       endif
--- a/autoload/deoplete/handlers.vim
+++ b/autoload/deoplete/handlers.vim
@@ -79,6 +79,10 @@ function! s:on_insert_leave() abort "{{{
 endfunction"}}}
 
 function! s:complete_done() abort "{{{
+  if exists('g:deoplete#_context.saved_completeopt')
+    let &completeopt = g:deoplete#_context.saved_completeopt
+    unlet g:deoplete#_context.saved_completeopt
+  endif
   let g:deoplete#_context.position = getpos('.')
 endfunction"}}}
 
--- a/autoload/deoplete/init.vim
+++ b/autoload/deoplete/init.vim
@@ -49,7 +49,7 @@ function! deoplete#init#enable() abort "{{{
   endif
 
   if &completeopt !~# 'noinsert\|noselect'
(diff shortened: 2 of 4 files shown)


autoload/deoplete/init.vim
```diff
@@ -49,6 +49,7 @@ function! deoplete#init#enable() abort "{{{
   endif
 
   if &completeopt !~# 'noinsert\|noselect'
+    let l:save_completeopt = &completeopt
```
Shougo: Please don't use `l:` prefix.
`save_completeopt` is better.

Shougo:
I will create `b:deoplete` variable for it.
Please wait until I will implement.

Shougo:
@blueyed Ping

blueyed:
Pushed some changes.

I've removed this patch, not sure if it should be kept?
```
diff --git a/autoload/deoplete/handlers.vim b/autoload/deoplete/handlers.vim
index 1a46c41..6fb1704 100644
--- a/autoload/deoplete/handlers.vim
+++ b/autoload/deoplete/handlers.vim
@@ -80,6 +80,10 @@ function! s:on_insert_leave() abort "{{{
 endfunction"}}}
 
 function! s:complete_done() abort "{{{
+  if exists('b:deoplete_saved_completeopt')
+    let &completeopt = g:deoplete#_context.saved_completeopt
+    unlet g:deoplete#_context.saved_completeopt
+  endif
   let g:deoplete#_context.position = getpos('.')
 endfunction"}}}
 ```

File: autoload/deoplete/handlers.vim [hunk 2]
```diff
@@ -51,6 +51,8 @@ function! s:completion_begin(event) abort "{{{
   " Save the previous position
   let g:deoplete#_context.position = context.position
 
+  call deoplete#mappings#_set_completeopt()
```
Shougo: No.  You should not call the function in the position.
The function must be called before of `feedkeys()` or `complete()`.

Shougo:
> I've removed this patch, not sure if it should be kept?

You should keep the patch.

autoload/deoplete/handlers.vim
blueyed: I needed this change, because otherwise it would select and insert the first entry automatically.

I am using `completeopt=longest,menuone` by default.

I have `context.filetypes = ['python', 'python']` here, and the `feedkeys("\<C-x>\<C-o>", 'n')` does not get triggered.  And "python" is not in `g:deoplete#_omni_patterns`, which is `{'mkd': ['<[^>]*'], 'xhtml': ['<[^>]*'], 'go': ['[^. \t0-9]\.\w*'], 'markdown': ['<[^>]*'], '
xml': ['<[^>]*'], 'c': ['[^. \t0-9]\.\w*', '[^. \t0-9]->\w*'], 'html': ['<[^>]*'], 'java': ['
[^. \t0-9]\.\w*'], 'cpp': ['[^. \t0-9]\.\w*', '[^. \t0-9]->\w*', '[a-zA-Z_]\w*::\w*']}`.

So then `call rpcnotify(g:deoplete#_channel_id, 'completion_begin', context)` gets triggered, and s:completion_begin will call `let context = deoplete#init#_context(a:event, [])` - overwriting the saved completeopt.

I think the problem is that completeopt cannot be set in an `<c-r>=`-mapping just before calling `complete()`?! https://github.com/blueyed/deoplete.nvim/blob/6e8a4c8fa33f73a8bdbcd5e93453ea5809cb39af/autoload/deoplete/mappings.vim#L26-L34

blueyed: Pushed a fix for this in https://github.com/blueyed/deoplete.nvim/commit/69d645dbcfb86af6bf2fa0d4f3c2294f1a09e449.

autoload/deoplete/handlers.vim
```diff
@@ -72,10 +72,17 @@ function! s:completion_begin(event) abort "{{{
 endfunction"}}}
 
 function! s:on_insert_leave() abort "{{{
+  if exists('g:deoplete#_context.saved_completeopt')
+    let &completeopt = g:deoplete#_context.saved_completeopt
```
Shougo: You should add the unlet.

autoload/deoplete/mappings.vim
```diff
@@ -29,8 +29,6 @@ function! deoplete#mappings#_init() abort "{{{
 endfunction"}}}
 
 function! deoplete#mappings#_do_complete(context) abort "{{{
-  call deoplete#mappings#_set_completeopt()
-
   if b:changedtick == get(a:context, 'changedtick', -1)
```
Shougo: You should add `call deoplete#mappings#_set_completeopt()
-` here.

rplugin/python3/deoplete/__init__.py
```diff
@@ -1,4 +1,4 @@
-#=============================================================================
+# ============================================================================
```
Shougo: Why did you added the changes?
If it is needed for the coding style, please add in the all python files.

rplugin/python3/deoplete/__init__.py
```diff
@@ -63,7 +64,10 @@ def completion_begin(self, context):
         var_context['candidates'] = candidates
         self.vim.vars['deoplete#_context'] = var_context
 
+        # Set (and store) current &completeopt setting.  This cannot be done
+        # (currently) from the deoplete_start_complete mapping's function.
+        self.vim.command(
+            'call deoplete#mappings#_set_completeopt()')
```
Shougo: It must be called before `feedkeys()` instead of `complete()`?
I don't know the problem.  Please explain it.

Shougo:
The travis error is fixed.

autoload/deoplete/handlers.vim
blueyed: Although the context gets reset just below?

rplugin/python3/deoplete/__init__.py
blueyed: Yeah, coding style: it's a separate commit, and I've noticed then that there are too many, and stopped before installing `autopep8`.  I've not noticed that there are so many Python files in the beginning.

autoload/deoplete/mappings.vim
blueyed: I've moved it - see below for an answer.

autoload/deoplete/handlers.vim
Shougo: Yes.

rplugin/python3/deoplete/__init__.py
blueyed: https://github.com/Shougo/deoplete.nvim/pull/79#discussion_r46074466

rplugin/python3/deoplete/__init__.py
Shougo: OK, if you fix the problems, please change the other headers.

rplugin/python3/deoplete/__init__.py
Shougo: OK, it seems Vim's feature.

autoload/deoplete/mappings.vim
Shougo: OK.

rplugin/python3/deoplete/__init__.py
blueyed: Reverted it.

autoload/deoplete/handlers.vim
blueyed: Done.

Shougo:
It seems good.  Thanks!

blueyed:
Cool, although that I would have squashed it into single commit (or at least removed the add+revert and the "fixup!", etc ... ;)

Shougo:
Unfortunately, deoplete cannot restore `completefunc`.
Please see #107 problem.

blueyed:
That would be really bad.. there must be some way - even when that means doing another patch for Vim/Neovim.
What is needed for this to be done properly?

Shougo:
> What is needed for this to be done properly?

The completion is called automatically, so we cannot decide the restore timing.

https://github.com/Valloric/YouCompleteMe/blob/07f4402f49a6cb987ebb17a4c17790816e06c3e7/autoload/youcompleteme.vim#L361

YouCompleteMe changes the completeopt and don't restore it.

blueyed:
FWIW, I had submitted a PR and moved it to my fork for YCM, where it can also be restored.

I am not seeing the issue reported in #107 myself, but have not tried the minimal configuration.
IIRC it will be restored on leaving insert mode, which should not interfere with automatic completion?!

Is this a problem maybe when `CompleteDone` is not supported by an older Vim?

Shougo:
> I am not seeing the issue reported in #107 myself, but have not tried the minimal configuration.

You must use the minimal configuration and html omni completion.

> Is this a problem maybe when CompleteDone is not supported by an older Vim?

No `CompleteDone` works.  But it works after the completion.
So it conflicts the auto completion behavior.

blueyed:
@Shougo 
> No `CompleteDone` works. But it works after the completion.
So it conflicts the auto completion behavior.

The restoring is especially meant for when you toggle the autocompletion manually (and trigger it manually, too).

Would it be feasible to have a function for enabling and disabling the auto completion behavior, where this would be (re)stored?

Or what about not restoring it on `CompleteDone` actually?! (https://github.com/Shougo/deoplete.nvim/pull/79#issuecomment-160135677)

blueyed:
A new attempt: https://github.com/Shougo/deoplete.nvim/pull/202

PR #86, opened 2015-11-30
doc: fix code selection marker from d725770

blueyed:


blueyed:
For some reason I am seeing the `<` and `>` markers only in insert mode..

Shougo:
Thanks.

blueyed:
JFI:

> For some reason I am seeing the < and > markers only in insert mode..

That's because they are concealed, and with the default of `concealcursor=nv` they can be made visible by visual selection also.  Another way is to set `conceallevel=0` to show them.

PR #88, opened 2015-12-04
WIP: Building a source for Jedi

ViViDboarder:
I don't actually intend for you to merge this into deoplete, but this was the easiest way to develop, debug, and ask for some help.

I've been working on the following source and I can see that the debug on line 11 works just fine, but it does not appear that `gather_candidates` is called at all because line 18 is never shown in messages. Is there something I'm missing?

Also, I'm getting the cursor and buffer contents from `self.vim.current`, but I see that a `context` is provided. Are these included? I would just read my debugs, but as I said before, `gather_candidates` doesn't appear to be called.

### DIFF
--- a/rplugin/python3/deoplete/sources/deoplete_jedi.py
+++ b/rplugin/python3/deoplete/sources/deoplete_jedi.py
@@ -1,0 +1,47 @@
+from jedi import Script
+
+from .base import Base
+from deoplete import util
+
+
+class Source(Base):
+    def __init__(self, vim):
+        # super(Base, self).__init__(vim)
+        Base.__init__(self, vim)
+        util.debug(vim, 'Init deoplete jedi')
+
+        self.name = 'jedi'
+        self.mark = '[J]'
+        self.filetypes = ['python']
+
+    def gather_candidates(self, context):
+        util.debug(self.vim, 'Init deoplete jedi')
+
+        util.debug(self.vim, context['input'])
+        util.debug(self.vim, context['complete_pos'])
+        util.debug(self.vim, context['complete_str'])
+        util.debug(self.vim, context['filetype'])
+        util.debug(self.vim, context['filetypes'])
+
+        row, column = self.vim.current.window.cursor
+        # If saved...
+        buf = self.vim.current.buffer
+        source = '\n'.join(buf[:])
+        path = buf.name
+        encoding = buf.options['fileencoding']
+
+        script = Script(
+            source=source,
+            row=row,
+            column=column,
+            path=path,
+            encoding=encoding,
+        )
+
+        return [
+            {
+                'word': completion.complete,
+                'kind': completion.type.full_name,
+            }
+            for completion in script.completions()
+        ]
--- a/rplugin/python3/deoplete/sources/jedi.py
+++ b/rplugin/python3/deoplete/sources/jedi.py
@@ -13,6 +13,7 @@ class Source(Base):
         self.name = 'jedi'
         self.mark = '[J]'
         self.filetypes = ['python']
+        self.min_pattern_length = 0
 
     def gather_candidates(self, context):
         util.debug(self.vim, 'Init deoplete jedi')


rplugin/python3/deoplete/sources/deoplete_jedi.py
```diff
@@ -0,0 +1,47 @@
...
+
+from .base import Base
+from deoplete import util
+
+
+class Source(Base):
+    def __init__(self, vim):
+        # super(Base, self).__init__(vim)
+        Base.__init__(self, vim)
+        util.debug(vim, 'Init deoplete jedi')
```
Shougo: `debug()` is included in base class currently.

rplugin/python3/deoplete/sources/deoplete_jedi.py
```diff
@@ -0,0 +1,47 @@
...
+
+        self.name = 'jedi'
+        self.mark = '[J]'
+        self.filetypes = ['python']
```
Shougo: `self.min_pattern_length = 0` is needed.

Shougo:
I think "jedi.py" is better for the filename.

Shougo:
> I've been working on the following source and I can see that the debug on line 11 works just fine, but it does not appear that gather_candidates is called at all because line 18 is never shown in messages. Is there something I'm missing?

You should define `get_complete_position()` to define complete position.

> Also, I'm getting the cursor and buffer contents from self.vim.current, but I see that a context is provided. Are these included? I would just read my debugs, but as I said before, gather_candidates doesn't appear to be called.

You can use `context['position']` for it.

ViViDboarder:
Is `get_complete_position()` not optional? https://github.com/Shougo/deoplete.nvim/blob/master/doc/deoplete.txt#L475

I'll try your suggestions.

Shougo:
> Is get_complete_position() not optional? https://github.com/Shougo/deoplete.nvim/blob/master/doc/deoplete.txt#L475

Yes, but it is not for omni completions like jedi.py.

ViViDboarder:
So I'm looking at the examples of `get_complete_position()` and I'm not sure the reasoning as to why jedi would require something special to determine the completion instead of defaulting to keyword.

Maybe I don't fully understand what the method is intended to do and what I should be doing with it.

Would it make sense for me to extend the `omni.Source` rather than `base.Base`? Would it be required to have much of the same implementation?

Shougo:
> So I'm looking at the examples of get_complete_position() and I'm not sure the reasoning as to why jedi would require something special to determine the completion instead of defaulting to keyword.

Do you know the omnifunc implementation in Vim?
omnifunc must return the complete position.

> Would it make sense for me to extend the omni.Source rather than base.Base? Would it be required to have much of the same implementation?

base is just default implementation.  But it is not for the omnifunc.

wsdjeg:
in omnifunc ,the complete func must be call twice，first check context and return the start and second return a complete list

ViViDboarder:
I'm having some trouble implementing `get_complete_position()`. What position should I be returning? How is it different than `context['position'][2]`?

And, somewhat related, why are there 4 indexes in the position? It seems like the first and last are always 0 for me.

wsdjeg:
as i known vim has two type list for a position,one has two indexes and the other has four,but now i am not near my pc,:h search :h getpos maybe help you

ViViDboarder:
@wsdjeg ah, thanks! That answers one of my questions. So that's probably what the other indexes are.
```
getpos({expr})  Get the position for {expr}.  For possible values of {expr}
                see line().  For getting the cursor position see
                getcurpos().
                The result is a List with four numbers:
                    [bufnum, lnum, col, off]
```

wsdjeg:
the first index is the number of the buffer,by default the last one is off or 0,that means the lnum and col are counted from the first chars which is not a space

wsdjeg:
why jedi can not use omni.vim source

ViViDboarder:
@wsdjeg ¯\_(ツ)_/¯

I suppose I could. I thought there may be some additional benefit from writing the plugin in pure python since the omni version is Python3 (deoplete) calling jedi-vim (vimscript) which is then calling Python2 (jedi module).

This implementation seems necessarily clunky. If I could create a pure Python3 implementation, it may be better off.

@Shougo does this reasoning make sense? Or should I just use omni?

Shougo:
Yes, we can use omni source with jedi.vim.
But it has some problems.

* If call omnifunc, Vim stops for a while.  Because, Vim(omnifunc) is not asynchronous.

* jedi.vim has unnecessary features for deoplete.  For example, built-in completion

* YouCompleteMe does not use jedi.vim.  It calls jedi directly.

I think it is to create deoplete source is better if possible.

zchee:
@ViViDboarder I have developed [deoplete-jedi](https://github.com/zchee/deoplete-jedi) with @Shougo  assistance.

I was very reference to your PR. Thanks!!
Could you try test it?

ViViDboarder:
@zchee awesome! I'll check it out.

zchee:
@ViViDboarder Thanks :)

Shougo:
I think deoplete-jedi is better.  Closing.

PR #105, opened 2015-12-24
Add TypeScript to JS omni-patterns list

indiesquidge:
This allows for omni-completion on object properties for TypeScript. The feature already existed for JavaScript, this just extends that functionality to TypeScript.


Shougo:
Thanks.

PR #106, opened 2015-12-26
Update readme

indiesquidge:
This PR updates the wording in the README so as to make it clear that there are _some_ customizations for this plugin, and if they would like to read about them they can check out the documentation.

This was initially discussed in this [issue](https://github.com/neovim/neovim/issues/1512#issuecomment-167271183).


Shougo:
OK, thanks.

Issue #126, opened 2016-01-14
Feature - disable completion for specified highlight styles


brendanjerwin:
I'd like to disable completion (or perhaps just change sources to just buffer or something) when the cursor is within certain kinds of text.

I often get unusable completions while working in strings or comments.

Shougo:
I don't think it is useful.
Because, highlight may be wrong and you can use `eval()` like functions.

brendanjerwin:
OK. I clearly have much more to learn. :)

PR #127, opened 2016-01-14
makes the popup respect the disable flags

brendanjerwin:
While attempting to disable deoplete in some cases I found that setting the disable flags didn't actually prevent the popupmenu.


brendanjerwin:
Useful for #126 if one wants to do so.

Shougo:
OK.  I improved it.

PR #171, opened 2016-02-14
Added a debug mode to prevent uncommented debug statements

nhooyr:
If debug is called but debug mode isn't set, an error is displayed. This
should help prevent uncommented debug statements.


Shougo:
Thanks.

PR #187, opened 2016-02-23
Fix outdated function name in doc

jebaum:


zchee:
@jebaum Thank PR!

Good catch.
`Examples` seems is correct.


/cc @Shougo

Shougo:
Thank you!

PR #202, opened 2016-03-08
Save 'completeopt' and restore it when leaving insert mode

blueyed:
This follows up on https://github.com/Shougo/deoplete.nvim/pull/79, but
does only restore it when leaving insert mode (not on CompleteDone).

This does not use `g:deoplete#_context` for it, because that would
require to pass it through in `completion_begin` and keep it in
`deoplete#deoplete#init#_context`.


Shougo:
OK.  Merged.

blueyed:
Thanks!

PR #203, opened 2016-03-08
Improve tag source: doc and filtering by keyword_patterns

blueyed:
This filters the tags by the current `keyword_patterns`.

It also changes the regexp for
`deoplete.sources.base.get_complete_position` to use a non-capturing
group.

### DIFF
--- a/rplugin/python3/deoplete/sources/base.py
+++ b/rplugin/python3/deoplete/sources/base.py
@@ -47,7 +47,7 @@ class Base(object):
         self.disabled_syntaxes = []
 
     def get_complete_position(self, context):
-        m = re.search('(' + context['keyword_patterns'] + ')$',
+        m = re.search('(?:' + context['keyword_patterns'] + ')$',
                       context['input'])
         return m.start() if m else -1
 
--- a/rplugin/python3/deoplete/sources/tag.py
+++ b/rplugin/python3/deoplete/sources/tag.py
@@ -24,9 +24,12 @@
 # }}}
 # ============================================================================
 
-from os.path import getmtime, exists, getsize
+import re
 from collections import namedtuple
+from os.path import exists, getmtime, getsize
+
 from deoplete.util import parse_file_pattern
+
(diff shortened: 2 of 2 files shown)


doc/deoplete.txt
```diff
@@ -429,6 +429,8 @@ member						*deoplete-source-member*
 tag						*deoplete-source-tag*
 		It collects keywords from |ctags| files.
 		Note: It only supports UTF-8 encoding tag file.
+		Note: https://github.com/Shougo/neoinclude.vim is required for
+		this.
```
Shougo: The change is wrong.
neoinclude is not required!
If you want to complete the candidates from included files, neoinclude is required.

doc/deoplete.txt
```diff
@@ -769,7 +771,8 @@ It prints the documentation you have completed.
 https://github.com/Shougo/echodoc.vim
 
 neoinclude.vim:
-You can completes the candidates from the included files and included path.
+You can complete the candidates from the included files and included paths.
+This is also required for |deoplete-source-tag|.
```
Shougo: It is not required.

Shougo:
Please describe your changes.
I don't want to merge no description PR.

doc/deoplete.txt
blueyed: Oh.. but I was not seeing tag sources (at some time), and when digging into the source have seen the reference to neoinclude.
Need to change that / investigate then.

blueyed:
@Shougo 
Updated, and removed the doc changes.

Shougo:
Thanks.

Why this change is needed?
Please describe the reason.

blueyed:
Added to the commit message:

> All tags are read and cached from the tags file, but then the source
will only return the matching/relevant ones.
This will include e.g. `deoplete#foo` in a `ft=vim` file, but not with
`ft=html`.

Shougo:
OK.  Thanks.

Issue #204, opened 2016-03-08
Add option for delayed auto-completion


blueyed:
Firing events on every keypress via `TextChangedI` is very performance intense, and I would rather like to have automatic completion with a small delay (e.g. 500ms).

In jedi-vim I've added the following, which uses a hack through CursorHoldI.
(There might be a better mechanism for Neovim (?), but the idea stays the same.)

    if g:jedi#show_call_signatures_delay > 0
        autocmd InsertEnter <buffer> let b:_jedi_orig_updatetime = &updatetime
                    \ | let &updatetime = g:jedi#show_call_signatures_delay
        autocmd InsertLeave <buffer> if exists('b:_jedi_orig_updatetime')
                    \ |   let &updatetime = b:_jedi_orig_updatetime
                    \ |   unlet b:_jedi_orig_updatetime
                    \ | endif
        autocmd CursorHoldI <buffer> call jedi#show_call_signatures()
    else
        autocmd CursorMovedI <buffer> call jedi#show_call_signatures()
    endif

What do you think about it?

MichaelMackus:
:+1: anything to improve the autocompletion performance. I notice when typing text in neovim with deoplete autocomplete enabled, long text strings take very long to appear into the buffer.

blueyed:
@Shougo 
It would be great to have your input before I start hacking on it.

blueyed:
@MichaelMackus 
You could give #210 a try.

MichaelMackus:
Seems to work nicely for me, thanks @blueyed ! One thing that would make the PR perfect for me is for the "deoplete_omni_patterns" to automatically popup the omnifunc automatically (instead of waiting for the delay) when the pattern is matched, but not a big deal at all.

Hopefully we can get this merged into master because it at least makes deoplete's performance bearable.

blueyed:
> "deoplete_omni_patterns" to automatically popup the omnifunc automatically

Good point!

blueyed:
@MichaelMackus 
Another attempt: #212.

Shougo:
I have implemented the feature.

PR #210, opened 2016-03-09
WIP: Add auto_complete_delay setting (defaulting to 500ms)

blueyed:
Implements: https://github.com/Shougo/deoplete.nvim/issues/204.

According to @justinmk it could use a timer in Python, but still needs to call Python then to trigger it (initially).  And it's not clear to me how to do this best.

### DIFF
--- a/autoload/deoplete/handlers.vim
+++ b/autoload/deoplete/handlers.vim
@@ -29,8 +29,21 @@ function! deoplete#handlers#_init() abort "{{{
     autocmd CompleteDone * call s:complete_done()
     autocmd InsertCharPre * call s:on_insert_char_pre()
 
-    autocmd TextChangedI * call s:completion_begin("TextChangedI")
     autocmd InsertEnter * call s:completion_begin("InsertEnter")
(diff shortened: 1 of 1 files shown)


Shougo:
This is not I thought the implementation.

tweekmonster:
@Shougo do you think you could reconsider even if it's a hard coded delay of 100ms or maybe a way to cancel previously triggered completions that are still pending?  I think a short delay would improve the responsiveness and avoid unnecessary completions.  I commented on deoplete-jedi about a problem that I thought this PR would help with: https://github.com/zchee/deoplete-jedi/issues/13#issuecomment-194684768

To summarize: Jedi is slow with broad completion searches and typing fast will cause deoplete-jedi to block subsequent completions until the previous ones are completed.

justinmk:
@tweekmonster https://github.com/Shougo/deoplete.nvim/pull/212

Shougo:
Yes, this feature is not rejected.

tweekmonster:
Ah sorry. I misread. Thanks!

PR #213, opened 2016-03-11
Revert remove empty deoplete.py

zchee:
Revert https://github.com/Shougo/deoplete.nvim/pull/211
See https://github.com/Shougo/deoplete.nvim/commit/5b6cb7c92bed7c88a6742b8ec603ee17710f1b51#commitcomment-16638255

/cc @blueyed


zchee:
This is contrary to the spirit of the python, but It is necessary to modify the implementation of neovim. Merge.

bfredl:
If think this is the offending line: https://github.com/neovim/python-client/blob/master/neovim/__init__.py#L40 . We need to change it to either accept `.py` or directory with `__init__.py`

zchee:
@bfedl Thanks for advice!!

Yes, this problem caused by your suggestion implements.
but I do not yet understand behavior of the Python core :(

Do you have any plan for resolve of this feature?

bfredl:
yes: https://github.com/neovim/python-client/pull/183

zchee:
@bfredl Oh! Thanks for fast response and resolve plan :)
I look forward to this feature.

zchee:
@bfredl I tested remove empty `deoplete.py` with local merging https://github.com/neovim/python-client/pull/183 and https://github.com/neovim/neovim/pull/4438. It worked perfectly :)

Because it seems to have backward compatibility, So if has been merged their PR, I (or @Shougo) will re-remove `deoplete.py`.
Thank!

Issue #206, opened 2016-03-09
autocompletion automatically inserts (wrong) text


lewicki-pk:
# Problems summary
When I'm writing my C++ code autocompletion automatically inserts first value from the list.

## Expected
I would like to just see the autocompletion list and disable option responsible for inserting actual characters from autocompletion menu.

## Environment Information

 * OS: Ubuntu 15.10
 * Neovim version: NVIM 0.1.3-dev

## Provide a minimal init.vim with less than 50 lines (Required!)

```vim
call plug#begin('~/.nvim/plugged')

Plug 'scrooloose/nerdtree'
Plug 'Shougo/deoplete.nvim'
Plug 'zchee/deoplete-clang'
Plug 'Shougo/neoinclude.vim'
Plug 'benekastah/neomake'
Plug 'bling/vim-airline'
Plug 'vim-airline/vim-airline-themes'
Plug 'jistr/vim-nerdtree-tabs'
Plug 'majutsushi/tagbar'
Plug 'dantler/vim-alternate'
Plug 'ctrlpvim/ctrlp.vim'
Plug 'szw/vim-tags'

call plug#end()

let g:deoplete#enable_at_startup = 1

set completeopt-=preview

let g:deoplete#sources#clang#libclang_path = '/usr/lib/llvm-3.6/lib/libclang.so.1'
let g:deoplete#sources#clang#clang_header = '/usr/lib/clang'
let g:deoplete#sources#clang#std#c = 'c11'
let g:deoplete#sources#clang#std#cpp = 'c++1z'
let g:deoplete#sources#clang#sort_algo = 'priority'

```

## The reproduce ways from neovim starting (Required!)

Currently I cannot think of a simple way to reproduce this. I came across this during following tutorial for cppCMS library (but I hope it's some minor issue or option that I simply don't know).

## Generate a logfile if appropriate

 1. export NVIM_PYTHON_LOG_FILE=/tmp/log
 2. export NVIM_PYTHON_LOG_LEVEL=DEBUG
 3. nvim -u minimal.vimrc
 4. some works
 5. cat /tmp/log_{PID}

## Screen shot (if possible)
https://www.youtube.com/watch?v=AaHg1IZ4890&feature=youtu.be

## Upload the log file
[log.zip](https://github.com/Shougo/deoplete.nvim/files/165173/log.zip)

zchee:
@lewicki-pk Hi :)
This problem seem to deoplete-clang isuue.
[...]

PR #219, opened 2016-03-12
Do not call 'deoplete#mappings#_set_completeopt' in manual mode

blueyed:
Changing `'completeopt'` seems to be necessary only for the
auto-completion mode?!

### DIFF
--- a/rplugin/python3/deoplete/deoplete.py
+++ b/rplugin/python3/deoplete/deoplete.py
@@ -79,9 +79,12 @@ class Deoplete(object):
             'event': context['event'],
         }
 
-        # Set (and store) current &completeopt setting.  This cannot be done
-        # (currently) from the deoplete_start_complete mapping's function.
-        self.__vim.call('deoplete#mappings#_set_completeopt')
+        if context['event'] != 'Manual':
+            # Set (and store) current &completeopt setting.  This cannot be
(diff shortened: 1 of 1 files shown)


Shougo:
I don't think so.  Closing.

blueyed:
I don't get it, please explain.

The manual completion is invoked through a user's script, which then can ensure the proper setup for `'completeopt'`.

With this patch/PR it's easier to call the manual completion, without having to (re)store the `'completeopt'` setting.

As far as I understood it, the changing of `'completeopt'` is necessary to work in the context of automatic completion, to not interfere with the user's typing process.  But in the case of manually triggering it this should not be required?!

Shougo:
I can implement it more properly.
But I don't want to add the restore feature.
Because, it may conflict the user config or user settings.

Example: https://github.com/Shougo/deoplete.nvim/issues/206

It seems the restore feature is broken.

Shougo:
If the feature is really needed, other needed people is needed.

blueyed:
Is #206 really related?  According to https://github.com/Shougo/deoplete.nvim/issues/206#issuecomment-196324186 it's not?!

This PR / suggestion does not have a (new) restore feature, but keeps using the existing one.
Do you mean to remove that one (again)?

Shougo:
I have implemented it.

blueyed:
Thanks!

Issue #218, opened 2016-03-12
Completion not triggered on first "InsertEnter"


blueyed:
When you enter insert mode right after starting Vim, before `CursorHold` is triggered, the completion will only get setup, but not triggered:

    if get(g:, 'deoplete#enable_at_startup', 0)
      augroup deoplete
        autocmd CursorHold,InsertEnter * call deoplete#init#enable()
      augroup END

In this case (for "InsertEnter") the completion_begin method should get called after init.

Similar to this, except for that `s:completion_begin` is not available in that scope:

```diff
diff --git i/autoload/deoplete/init.vim w/autoload/deoplete/init.vim
index 945ca7a..449d634 100644
--- i/autoload/deoplete/init.vim
+++ w/autoload/deoplete/init.vim
@@ -31,10 +31,11 @@ function! deoplete#init#is_enabled() abort "{{{
   return s:is_enabled
 endfunction"}}}
 
-function! deoplete#init#enable() abort "{{{
+function! deoplete#init#enable(...) abort "{{{
   if deoplete#init#is_enabled()
     return
   endif
+  let event = a:0 ? a:1 : ''
 
   augroup deoplete
     autocmd!
@@ -78,6 +79,10 @@ function! deoplete#init#enable() abort "{{{
   call deoplete#init#_variables()
   call deoplete#handlers#_init()
   call deoplete#mappings#_init()
+
+  if event == 'InsertEnter'
+    call s:completion_begin("InsertEnter")
+  endif
 endfunction"}}}
 
 function! deoplete#init#_variables() abort "{{{
diff --git i/plugin/deoplete.vim w/plugin/deoplete.vim
index e453657..bdad8f9 100644
--- i/plugin/deoplete.vim
+++ w/plugin/deoplete.vim
@@ -34,7 +34,8 @@ command! -nargs=0 -bar DeopleteEnable
 " Global options definition. "{{{
 if get(g:, 'deoplete#enable_at_startup', 0)
   augroup deoplete
-    autocmd CursorHold,InsertEnter * call deoplete#init#enable()
+    autocmd CursorHold * call deoplete#init#enable()
+    autocmd InsertEnter * call deoplete#init#enable("InsertEnter")
   augroup END
 endif
 "}}}
```

Shougo:
Closing.

blueyed:
@Shougo 
[...]

PR #221, opened 2016-03-13
Fix init through CursorHold

blueyed:
Really fixes https://github.com/Shougo/deoplete.nvim/issues/218.

### DIFF
--- a/plugin/deoplete.vim
+++ b/plugin/deoplete.vim
@@ -28,13 +28,14 @@ if exists('g:loaded_deoplete')
 endif
 let g:loaded_deoplete = 1
 
-command! -nargs=0 -bar DeopleteEnable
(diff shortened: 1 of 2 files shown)


plugin/deoplete.vim
```diff
@@ -28,13 +28,14 @@ if exists('g:loaded_deoplete')
 endif
 let g:loaded_deoplete = 1
 
-command! -nargs=0 -bar DeopleteEnable
-      \ call deoplete#init#enable()
+command! -nargs=0 -bar DeopleteEnable call deoplete#init#enable()
```
zchee: Why do you have a fix of indent that does not change the behavior?

Shougo:
OK.

PR #222, opened 2016-03-14
Fix always call vim.vars even if setting is false

zchee:
Always get `g:deoplete#enable_profile` value if not set it flag
- ~~Also gives priority to the `g:deoplete#enable_debug` flag ahead~~

and support only once to get the this flag.

Before and after logs here.

- before: https://gist.github.com/zchee/681c113554f5022ce94d#file-before-log
- after: https://gist.github.com/zchee/681c113554f5022ce94d#file-after-log

===

Edit:
- Parse `g:deoplete#enable_profile` only. `g:deoplete#enable_debug` for debug echomsg.


Shougo:
Thanks.

PR #225, opened 2016-03-16
Fix debug when args is not a json

zchee:
Fix if args is not json syntax.
and `strip()` for `cont['input']` or etc.


Shougo:
Thanks.

PR #228, opened 2016-03-17
Use `.endswith` with disabled_syntaxes check

blueyed:
It seems to be a common habit to also ignore `gitcommitComment` for
`Comment`?!

Otherwise, we could have support for globbing here, i.e. only `*Comment`
would also ignore `gitcommitComment`.

### DIFF
--- a/rplugin/python3/deoplete/deoplete.py
+++ b/rplugin/python3/deoplete/deoplete.py
@@ -311,8 +311,8 @@ class Deoplete(object):
                        not (min_pattern_length <=
                             len(context['complete_str']) <=
                             max_pattern_length))
-        return (disabled_syntaxes and
-                context['syntax_name'] in disabled_syntaxes) or skip_length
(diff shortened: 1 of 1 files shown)


Shougo:
This is not needed.
"gitcommitComment" is converted to "Comment".  Closing.

blueyed:
Where is it converted?

Because I've noticed that the following did not appear to work?! (`gitcommitComment` was still handled)

> call deoplete#custom#set('_', 'disabled_syntaxes', ['Comment', 'String'])

Shougo:
> Where is it converted?

Please see `deoplete#util#get_syn_name()`.

Shougo:
https://github.com/vim/vim/blob/089af18d1fd0961ff504ee72db0156bbfe509cdf/runtime/syntax/gitcommit.vim#L57

`gitcommitComment` links to `Comment`.
So, `deoplete#util#get_syn_name()` to `Comment`.

PR #230, opened 2016-03-17
Use Python's logging facility

tweekmonster:
This uses Python's logging facility to log messages to a file instead of printing messages to Neovim's message history.  It produces the following output when `g:deoplete#enable_debug` is on:

```
2016-03-17 17:20:59,010 INFO     (deoplete.logging) --- Deoplete Log Start ---
2016-03-17 17:20:59,011 INFO     (deoplete.logging) NVIM 0.1.3-dev, Python 3.4.3, neovim client 0.1.4
2016-03-17 17:20:59,016 DEBUG    (deoplete.core) Loaded Source: ultisnips (/home/tallen/.vim/plugged/ultisnips/rplugin/python3/deoplete/sources/ultisnips.py)
2016-03-17 17:20:59,016 DEBUG    (deoplete.core) Loaded Source: buffer (/home/tallen/dev/vim/deoplete.nvim/rplugin/python3/deoplete/sources/buffer.py)
2016-03-17 17:20:59,017 DEBUG    (deoplete.core) Loaded Source: dictionary (/home/tallen/dev/vim/deoplete.nvim/rplugin/python3/deoplete/sources/dictionary.py)
2016-03-17 17:20:59,017 DEBUG    (deoplete.core) Loaded Source: file (/home/tallen/dev/vim/deoplete.nvim/rplugin/python3/deoplete/sources/file.py)
2016-03-17 17:20:59,018 DEBUG    (deoplete.core) Loaded Source: member (/home/tallen/dev/vim/deoplete.nvim/rplugin/python3/deoplete/sources/member.py)
2016-03-17 17:20:59,018 DEBUG    (deoplete.core) Loaded Source: omni (/home/tallen/dev/vim/deoplete.nvim/rplugin/python3/deoplete/sources/omni.py)
2016-03-17 17:20:59,019 DEBUG    (deoplete.core) Loaded Source: tag (/home/tallen/dev/vim/deoplete.nvim/rplugin/python3/deoplete/sources/tag.py)
2016-03-17 17:20:59,094 DEBUG    (deoplete.core) Loaded Source: deoplete_jedi (/home/tallen/dev/vim/deoplete-jedi/rplugin/python3/deoplete/sources/deoplete_jedi.py)
2016-03-17 17:20:59,098 DEBUG    (deoplete.core) Loaded Filter: converter_auto_delimiter (/home/tallen/dev/vim/deoplete.nvim/rplugin/python3/deoplete/filters/converter_auto_delimiter.py)
2016-03-17 17:20:59,098 DEBUG    (deoplete.core) Loaded Filter: converter_auto_paren (/home/tallen/dev/vim/deoplete.nvim/rplugin/python3/deoplete/filters/converter_auto_paren.py)
2016-03-17 17:20:59,098 DEBUG    (deoplete.core) Loaded Filter: converter_remove_overlap (/home/tallen/dev/vim/deoplete.nvim/rplugin/python3/deoplete/filters/converter_remove_overlap.py)
2016-03-17 17:20:59,098 DEBUG    (deoplete.core) Loaded Filter: converter_remove_paren (/home/tallen/dev/vim/deoplete.nvim/rplugin/python3/deoplete/filters/converter_remove_paren.py)
2016-03-17 17:20:59,099 DEBUG    (deoplete.core) Loaded Filter: matcher_full_fuzzy (/home/tallen/dev/vim/deoplete.nvim/rplugin/python3/deoplete/filters/matcher_full_fuzzy.py)
2016-03-17 17:20:59,099 DEBUG    (deoplete.core) Loaded Filter: matcher_fuzzy (/home/tallen/dev/vim/deoplete.nvim/rplugin/python3/deoplete/filters/matcher_fuzzy.py)
2016-03-17 17:20:59,099 DEBUG    (deoplete.core) Loaded Filter: matcher_head (/home/tallen/dev/vim/deoplete.nvim/rplugin/python3/deoplete/filters/matcher_head.py)
2016-03-17 17:20:59,099 DEBUG    (deoplete.core) Loaded Filter: matcher_length (/home/tallen/dev/vim/deoplete.nvim/rplugin/python3/deoplete/filters/matcher_length.py)
2016-03-17 17:20:59,099 DEBUG    (deoplete.core) Loaded Filter: sorter_rank (/home/tallen/dev/vim/deoplete.nvim/rplugin/python3/deoplete/filters/sorter_rank.py)
2016-03-17 17:20:59,859 WARNING  (deoplete.jedi) caching with key: /home/tallen/dev/vim/deoplete.nvim/rplugin/python3/deoplete/deoplete.py, lines: [48, 49]
2016-03-17 17:20:59,919 DEBUG    (deoplete.jedi) cache hit
2016-03-17 17:21:00,142 WARNING  (deoplete.jedi) caching with key: os, lines: [0, 0]
2016-03-17 17:21:05,350 WARNING  (deoplete.jedi) caching with key: os.access, lines: [0, 0]
2016-03-17 17:21:08,839 DEBUG    (deoplete.jedi) cache hit
2016-03-17 17:21:13,722 WARNING  (deoplete.jedi) caching with key: /home/tallen/dev/vim/deoplete.nvim/rplugin/python3/deoplete/deoplete.py.self, lines: [48, 49]
```

When it's disabled, it uses Python's `logging.NullHandler` so the overhead should be minimal while allowing you to keep debug messages in the source without needing to comment them out.

I set it so that if a logging source produces more than 2 error messages in succession, it is silenced and can't print to Neovim's messages until restarted.  It is still able to continue logging to the output file.  Since this plugin can potentially produce log messages on each key stroke, I also set it so that if the same message comes in within 0.5s of the last time it was logged, it is ignored.

The root logger 'deoplete' is set to not propagate, so that Neovim's logs can be inspected separately without noise from this logger.

### DIFF
--- a/doc/deoplete.txt
+++ b/doc/deoplete.txt
@@ -341,6 +341,14 @@ deoplete#custom#set({source-name}, {option-name}, {value})
 		call deoplete#custom#set('_',
 		\ 'disabled_syntaxes', ['Comment', 'String'])
 <
+					*deoplete#util#enable_logging()*
+deoplete#util#enable_logging({level}, {logfile})
+		Enable logging for debugging purposes.
+		Set {level} to DEBUG, INFO, WARNING, ERROR, or CRITICAL.
+		{logfile} is the file where log messages are written.
+		Messages are appended to this file.  Each log session will
+		start with "--- Deoplete Log Start ---".
+
 ------------------------------------------------------------------------------
 KEY MAPPINGS 					*deoplete-key-mappings*
 
@@ -440,7 +448,7 @@ FILTERS						*deoplete-filters*
 
 					*deoplete-filter-matcher_default*
 Default matchers: ['matcher_length', 'matcher_fuzzy']
-	
+
 	You can change it by |deoplete#custom#set()|.
 
 					*deoplete-filter-sorter_default*
--- a/rplugin/python3/deoplete/__init__.py
+++ b/rplugin/python3/deoplete/__init__.py
@@ -6,6 +6,7 @@
 
 import neovim
 
+from deoplete import logger
 from deoplete.deoplete import Deoplete
 
 
@@ -19,6 +20,11 @@ class DeopleteHandlers(object):
     def init_python(self, args):
         self.__deoplete = Deoplete(self.__vim)
         self.__vim.vars['deoplete#_channel_id'] = self.__vim.channel_id
+
+    @neovim.rpc_export('deoplete_enable_logging', sync=True)
+    def enable_logging(self, level, logfile):
+        logger.setup(self.__vim, level, logfile)
+        self.__deoplete.debug_enabled = True
 
     @neovim.rpc_export('deoplete_auto_completion_begin')
     def completion_begin(self, context):
--- a/rplugin/python3/deoplete/deoplete.py
+++ b/rplugin/python3/deoplete/deoplete.py
@@ -11,6 +11,7 @@ from deoplete.util import \
 import deoplete.sources
 import deoplete.filters
 import deoplete.util
+from deoplete import logger
 
 import re
 import importlib.machinery
@@ -23,7 +24,7 @@ deoplete.sources  # silence pyflakes
 deoplete.filters  # silence pyflakes
 
 
-class Deoplete(object):
+class Deoplete(logger.LoggingMixin):
 
     def __init__(self, vim):
         self.__vim = vim
@@ -32,6 +33,7 @@ class Deoplete(object):
         self.__runtimepath = ''
         self.__profile_flag = None
         self.__profile_start = 0
+        self.__logname = 'core'
 
     def completion_begin(self, context):
         pos = self.__vim.current.window.cursor
@@ -222,9 +224,6 @@ class Deoplete(object):
             candidate['icase'] = 1
         return (complete_position, candidates)
 
-    def debug(self, expr):
-        deoplete.util.debug(self.__vim, expr)
-
     def profile_start(self, name):
         if self.__profile_flag is 0:
             return
@@ -284,6 +283,7 @@ class Deoplete(object):
                 'converters', source.converters)
 
             self.__sources[name] = source
+            self.debug('Loaded Source: %s (%s)', name, module.__file__)
         # self.debug(self.__sources)
 
     def load_filters(self):
@@ -298,6 +298,7 @@ class Deoplete(object):
                 'deoplete.filters.' + name, path).load_module()
             if hasattr(filter, 'Filter') and name not in self.__filters:
                 self.__filters[name] = filter.Filter(self.__vim)
+                self.debug('Loaded Filter: %s (%s)', name, filter.__file__)
         # self.debug(self.__filters)
 
     def is_skip(self, context, disabled_syntaxes,
--- a/autoload/deoplete/util.vim
+++ b/autoload/deoplete/util.vim
@@ -138,4 +138,22 @@ function! deoplete#util#get_syn_name() abort "{{{
         \          col('.')-1 : col('.'), 1)), 'name') : ''
 endfunction"}}}
 
+function! deoplete#util#neovim_version() abort "{{{
+  redir => v
+  silent version
+  redir END
+  return split(v, '\n')[0]
+endfunction"}}}
+
+function! deoplete#util#enable_logging(level, logfile) abort "{{{
+  if !deoplete#init#is_enabled()
+    " Enable to allow logging before completions start
+    call deoplete#init#enable()
+  endif
+
+  call rpcrequest(g:deoplete#_channel_id,
+        \ 'deoplete_enable_logging',
+        \ a:level,
+        \ a:logfile)
+endfunction"}}}
 " vim: foldmethod=marker
--- a/rplugin/python3/deoplete/filters/base.py
+++ b/rplugin/python3/deoplete/filters/base.py
@@ -5,10 +5,10 @@
 # ============================================================================
 
 from abc import abstractmethod
-import deoplete.util
+from deoplete.logger import LoggingMixin
 
 
-class Base(object):
+class Base(LoggingMixin):
 
     def __init__(self, vim):
         self.vim = vim
@@ -18,6 +18,3 @@ class Base(object):
     @abstractmethod
     def filter(self, context):
         pass
-
-    def debug(self, expr):
-        deoplete.util.debug(self.vim, expr)
--- a/rplugin/python3/deoplete/logger.py
+++ b/rplugin/python3/deoplete/logger.py
@@ -1,0 +1,145 @@
+# ============================================================================
+# FILE: logger.py
+# AUTHOR: Tommy Allen <tommy at esdf.io>
+# License: MIT license
+# ============================================================================
+
+import sys
+import time
+import logging
+from functools import wraps
+from collections import defaultdict
+
+import pkg_resources
+
+log_format = '%(asctime)s %(levelname)-8s (%(name)s) %(message)s'
+log_message_cooldown = 0.5
+
+root = logging.getLogger('deoplete')
+root.propagate = False
+init = False
+
+
+def getLogger(name):
+    """Get a logger that is a child of the 'root' logger.
+    """
+    return root.getChild(name)
+
+
+def setup(vim, level, output_file=None):
+    """Setup logging for Deoplete
+    """
+    global init
+    if init:
+        return
+    init = True
+
+    if output_file:
+        formatter = logging.Formatter(log_format)
+        handler = logging.FileHandler(filename=output_file)
+        handler.setFormatter(formatter)
+        handler.addFilter(DeopleteLogFilter(vim))
+        root.addHandler(handler)
+
+        level = str(level).upper()
+        if level not in ('DEBUG', 'INFO', 'WARN', 'WARNING', 'ERROR',
+                         'CRITICAL', 'FATAL'):
+            level = 'DEBUG'
+        root.setLevel(getattr(logging, level))
+
+        log = getLogger('logging')
+        log.info('--- Deoplete Log Start ---')
+        log.info('%s, Python %s, neovim client %s',
+                 vim.call('deoplete#util#neovim_version'),
+                 '.'.join(map(str, sys.version_info[:3])),
+                 pkg_resources.get_distribution('neovim').version)
+        vim.call('deoplete#util#print_warning', 'Logging to %s' % output_file)
+
+
+def logmethod(func):
+    """Decorator for setting up the logger in LoggingMixin subclasses.
+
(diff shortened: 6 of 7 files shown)


zchee:
@tweekmonster Thanks. I was discuss with @Shougo 
I will review instead of @Shougo later

autoload/deoplete/init.vim
```diff
@@ -105,6 +105,10 @@ function! deoplete#init#_variables() abort "{{{
   call deoplete#util#set_default(
         \ 'g:deoplete#enable_debug', 0)
   call deoplete#util#set_default(
+        \ 'g:deoplete#debug_log', '/tmp/deoplete.log')
```
zchee: `/tmp/deoplete.log` is not Windows compatible.
It's vim's `tempname()` and `deoeplete.log` file name better than it.

rplugin/python3/deoplete/__init__.py
```diff
@@ -41,6 +42,12 @@ def init_python(self, args):
 
     @neovim.rpc_export('deoplete_auto_completion_begin')
     def completion_begin(self, context):
+        if not logger.init:
```
zchee: Can not debugging in `deoplete_manual_completion_begin`.
So, I think better way is that method moves to into `self.__deoplete.completion_begin(context)`

https://github.com/tweekmonster/deoplete.nvim/blob/logging/rplugin%2Fpython3%2Fdeoplete%2Fdeoplete.py#L57

zchee:
@tweekmonster I was commented.
and, It conflicts to `master` now. Could you rebase it?

tweekmonster:
@zchee I made the fixes.

>So, I think better way is that method moves to into self.__deoplete.completion_begin(context)

I thought it would be better if logging was setup in `Deoplete.__init__()` but the debug_enabled variable is not available at that point.  I put the setup in `__init__.py` because I thought that was the only place that could guarantee that the variable is available.

>It's vim's tempname() better than it.

I think it's preferable to use `deoplete.log` as the filename so you won't need to keep changing the file if you're using `tail -f` in another window.  I set the default to `$TEMP/deoplete.log` in the Vim variable so that Python can replace `$TEMP/` with the platform-specific temp dir.

zchee:
@tweekmonster Thanks :)

Now Shougo is busy, can not comment on this PR.
but discussed on Japanese.
I will re-review later.

===

BTW, How to print the debug log at the source side?
`self.debug()`?

tweekmonster:
> BTW, How to print the debug log at the source side?  self.debug()?

For clarity:

- `LoggingMixin` can be subclassed by anything to add `.debug()`, `.info()`, etc.  The subclass only needs to set `self.name` and `self.debug_enabled`.  It is the subclass's responsibility to enable logging for itself.
- `LoggingMixin` is only a convenience for subclasses and will only setup on the first log message.
- Logging can be used directly with the requirement that the logger name starts with `deoplete.`:

```python
import logging

log = logging.getLogger('deoplete.custom')
log.debug('test')  # Will emit to the log file
log.error('test')  # Will emit to Vim
```

- `.error()`, `.exception()`, and `.critical()` will emit messages inside of Vim to inform the user of a problem.
- `.error()` and `.critical()` can be emitted twice in a row before it is silenced.  This is to avoid cluttering the message log.
- `.exception()` is silenced immediately after they are emitted since this should signal unrecoverable problems.  This is to avoid duplicate stack traces that make it difficult to see the start of the problem.
- `.error()`, `.exception()`, and `.critical()` are still emitted to the log file after they are silenced.

After writing this, I think it might be beneficial if `.error()`, `.exception()`, and `.critical()` is still sent to Vim even if debugging is disabled.  What do you think?

zchee:
@tweekmonster Thanks for detail of how to use.
It is easy to understand. I was a little understanding.

I'll try it, and comment later (maybe tomorrow).

rplugin/python3/deoplete/deoplete.py
```diff
@@ -53,6 +55,13 @@ def __init__(self, vim):
         self.__profile_start = 0
 
     def completion_begin(self, context):
+        if not logger.init:
+            log_file = None
+            if self.__vim.vars['deoplete#enable_debug']:
+                self.debug_enabled = True
+                log_file = self.__vim.vars['deoplete#debug_log']
+            logger.setup(self.__vim, log_file)
```
zchee: Why `logger.setup` is `g:deoplete#enable_debug` even false?

If
```diff
        if not logger.init:
            log_file = None
            if self.__vim.vars['deoplete#enable_debug']:
                self.debug_enabled = True
                log_file = self.__vim.vars['deoplete#debug_log']
-           logger.setup(self.__vim, log_file)
+               logger.setup(self.__vim, log_file)
```
It works the same.

Also, I think not necessary `root.addHandler(logging.NullHandler())`
https://github.com/Shougo/deoplete.nvim/pull/230/files#diff-fff55db87196854df902973d7b43a6dbR37

zchee:
@tweekmonster I want to also modify the `doc/deoplete.txt`.
Add `g:deoplete#debug_log`, `g:deoplete#debug_level` and fix `g:deoplete#enable_debug` behavior.
simple is OK.

and, Sorry for `master` is changed.
Could you re-rebase it?

rplugin/python3/deoplete/deoplete.py
tweekmonster: > Why logger.setup is g:deoplete#enable_debug even false?

Mainly to set `logger.init` to `True` since I think that the logger's state variable should be managed internally.  If `self.__vim.vars['deoplete#enable_debug']` is checked each time, that would cause a slowdown (I think).  Something needs to know that `enable_debug` was already checked.

I feel that once this is added, it will be the least updated file in the repo.  I wanted to keep changes to the core minimal.  That was also the reason why I originally placed it in `__init__.py`

> Also, I think not necessary root.addHandler(logging.NullHandler())

I have a habit of being explicit if the cost is low.  I have no problem with removing this if you think it has no benefit.

tweekmonster:
@zchee I'll rebase, but I'll have to make changes tomorrow.  I have to go to bed earlier tonight :zzz:

zchee:
@tweekmonster Sorry for a late reply.
I will post the review later.

rplugin/python3/deoplete/deoplete.py
zchee: @tweekmonster 
> If self.__vim.vars['deoplete#enable_debug'] is checked each time, that would cause a slowdown (I think). Something needs to know that enable_debug was already checked.

Yes, maybe will check each time.
Alternatively, Is there problems in `int` or `None` check for `self.debug_enabled = ***`?
such as 
https://github.com/Shougo/deoplete.nvim/blob/master/rplugin/python3/deoplete/deoplete.py#L247

doc/deoplete.txt
```diff
@@ -155,8 +155,21 @@ b:deoplete_disable_auto_complete
...
-		It enables deoplete debug mode.  The debug print feature is
-		enabled.
+		Enables debug logging.  Debug messages are logged to a file
+		using Python's logging facility.
+
+				*g:deoplete#debug_log*
+g:deoplete#debug_log
+		Sets the location of the debug log.
+
+		Default value: $TEMP/deoplete.log
```
zchee: I want to add,
```vim
If not set the $TEMP, Python will replace platform-specific temp dir.
```
Please change if the strange is English spelling.

rplugin/python3/deoplete/deoplete.py
tweekmonster: The reason I did this is because of `self.__vim.vars['deoplete#enable_debug']` availability.  I also wanted it to be checked only once.  When `Deoplete.__init__()` is called, `deoplete#enable_debug` is not available until `call _deoplete()` occurs in VimL.  The other issue is that an instance of `vim` needs to be passed to `setup()` so that it's able to print errors in Vim.  I think it looks strange because of the conditions that need to be met for how I wrote it.

Maybe it's better to have the logging filter check `deoplete#enable_debug` lazily when the first message is emitted?  This way, `setup()` can be called in `Deoplete.__init__()` without initial checks.  What do you think?

doc/deoplete.txt
tweekmonster: Will do.  I'll fix in the morning.

rplugin/python3/deoplete/deoplete.py
tweekmonster: @zchee I was looking closer at things.  How about adding an RPC export for `deoplete_enable_debug`?  This way, the logging can be enabled from VimL where the variable can be checked when `_deoplete()` is called.  `logging.setup()` can be called in the RPC function instead.  With this method, it would be possible to enable logging from a function in VimL without needing to restart.

zchee: Hm, interest.
I'll check it.

tweekmonster:
@zchee Added RPC export to enabling logging.  You can enable it with `deoplete#util#enable_logging('DEBUG', '/tmp/deoplete.log')` without needing to set `enable_debug` in vimrc.  I think this method is better since no checks are needed on the Python side to enable logging.  Also, logging can be enabled at run time so that messages can be captured as soon as a problem is noticed instead of trying to figure out how to recreate the problem after restarting nvim.

It's only a proof of concept right now.  If it's acceptable, it will need to be updated so that  `deoplete#util#enable_logging` can accept no arguments to use the default values.  Neovim's logging is also enabled when calling this function to show how it can work.  If that's a desired feature, I think it'll be better to set our own FileHandler on it to make it easier to locate the log file (without the PID).

> If not set the $TEMP, Python will replace platform-specific temp dir.

Didn't add this yet in case I need to revert.

tweekmonster:
@zchee Updated changes per our discussion.  Please review when you get a chance.

Shougo:
@tweekmonster Please add the LICENSE information header in logger.py.

tweekmonster:
@Shougo Sure.  Give me one moment.

tweekmonster:
Done: https://github.com/Shougo/deoplete.nvim/pull/230/commits/e3c925dc5db5e2b6b96a262f67ea9713940cdad1#diff-fff55db87196854df902973d7b43a6dbR1

tweekmonster:
@Shougo Sorry.  I didn't see the other commit where you separated the license file.  Did you want me to follow the format of the new file headers?

Shougo:
@tweekmonster Yes, please follow it.

@zchee LGTM?  Please check the code.

zchee:
@Shougo I test and check yesterday.
No problem.

LGTM.

Shougo:
OK.  Merged.

zchee:
@Shougo Thanks!

tweekmonster:
@Shougo @zchee Awesome, thanks!

Shougo:
Note:  I will add autoload/deoplete.vim for user functions.
The function name will be changed later.

tweekmonster:
Yes, that would make more sense.

Shougo:
I have renamed.

PR #237, opened 2016-03-24
Integrate profile to logging

zchee:
Integrate profile result to logging output.
and, `25` width formatting.

If you do not like, I'll fix the original format.

gist: https://gist.github.com/zchee/6e801a19a60d86e69251
image:

screenshot 2016-03-24 13 52 33

### DIFF
--- a/rplugin/python3/deoplete/deoplete.py
+++ b/rplugin/python3/deoplete/deoplete.py
@@ -33,7 +33,7 @@ class Deoplete(logger.LoggingMixin):
         self.__runtimepath = ''
         self.__profile_flag = None
         self.__profile_start = 0
-        self.__logname = 'core'
+        self.name = 'core'
 
     def completion_begin(self, context):
         pos = self.__vim.current.window.cursor
--- a/rplugin/python3/deoplete/deoplete.py
+++ b/rplugin/python3/deoplete/deoplete.py
@@ -225,7 +225,7 @@ class Deoplete(logger.LoggingMixin):
         return (complete_position, candidates)
 
     def profile_start(self, name):
-        if self.__profile_flag is 0:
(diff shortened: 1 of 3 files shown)


rplugin/python3/deoplete/deoplete.py
```diff
@@ -33,7 +33,7 @@ def __init__(self, vim):
         self.__runtimepath = ''
         self.__profile_flag = None
         self.__profile_start = 0
-        self.__logname = 'core'
```
tweekmonster: giphy

tweekmonster:
@zchee I think this may be better if profiling was moved to `LoggingMixin` so that source plugins can be profiled, too.  It could check `self.debug_enabled` so that profiling won't occur when there's no logging available.

rplugin/python3/deoplete/deoplete.py
```diff
@@ -234,15 +234,14 @@ def profile_start(self, name):
...
-                'echomsg \'profile start: {0}\''.format(name))
+            self.debug('Profile Start: {0}'.format(name))
             self.__profile_start = time.clock()
 
     def profile_end(self, name):
         if self.__profile_start:
-            self.__vim.command(
-                'echomsg \'profile end  : {0}, time={1}\''.format(
-                    name, time.clock() - self.__profile_start))
+            self.debug('Profile End  : {0:<25} {1}'.format(
```
tweekmonster: Maybe the Formatter should be set to use the newer format strings: https://docs.python.org/3.4/library/logging.html#logging.Formatter

So this could be simplified to:
```python
self.debug('Profile End  : {0:<25} time={:2.10f}', name, time.clock() - self.__profile_start)
```

Shougo:
LGTM.

PR #241, opened 2016-03-26
Ignore missing pkg_resources

### LINKED ISSUE #239: Dependency on setuptools
Hi!

Commit 7ce47a109c802ebd0030540e5ef4cc0e1fa7b1df added logger.py file. It relies on setuptools package facilities
```
import pkg_resources
```
So either installation requirements should be updated, or code should be refactored to get rid of this dependency.

tweekmonster:
Fixes #239


Shougo:
Nice.

PR #254, opened 2016-04-17
Fix #253 bufnr does not correlate with list index

### LINKED ISSUE #253: Change filename by NERDTree will cause deoplete throw errors
# Problems summary

If you change filename in neovim by using plugin like NERDTree, deoplete will throw errors once you enter INSERT mode.

If you cange filename by execute shell command `:!mv ...`, errors will not happen. It may because NERDTree will automatically update the current buffer but shell command not, this will confuse deoplete to maintain the buffer context I guess.

Sometimes switch git branch may also cause this problem but not always.

## Environment Information

 * OS: OS X 10.11.4
 * Neovim version: v0.1.3

## Provide a minimal init.vim with less than 50 lines (Required!)

```vim
" Your minimal init.vim
set runtimepath+=~/path/to/deoplete.nvim/
let g:deoplete#enable_at_startup = 1
```

## The reproduce ways from neovim starting (Required!)

 1. open a file in Neovim
 2. use NERDTree to change the filename
 3. enter Insert Mode
 4. error happens

## Generate a logfile if appropriate

Below is the `:messages` after error happens:

```
[deoplete]   File "/Users/nightire/.config/nvim/plugins/deoplete.nvim/rplugin/python3/deoplete/deoplete.py", line 47, in completion_begin
[deoplete]     complete_position, candidates = self.gather_candidates(context)
[deoplete]   File "/Users/nightire/.config/nvim/plugins/deoplete.nvim/rplugin/python3/deoplete/deoplete.py", line 85, in gather_candidates
[deoplete]     results = self.gather_results(context)
[deoplete]   File "/Users/nightire/.config/nvim/plugins/deoplete.nvim/rplugin/python3/deoplete/deoplete.py", line 158, in gather_results
[deoplete]     context['candidates'] = source.gather_candidates(context)
[deoplete]   File "/Users/nightire/.config/nvim/plugins/deoplete.nvim/rplugin/python3/deoplete/sources/buffer.py", line 24, in gather_candidates
[deoplete]     self.__make_cache(context, self.vim.current.buffer.number)
[deoplete]   File "/Users/nightire/.config/nvim/plugins/deoplete.nvim/rplugin/python3/deoplete/sources/buffer.py", line 59, in __make_cache
[deoplete]     p.findall(x) for x in self.vim.buffers[bufnr]
[deoplete]   File "/Users/nightire/Library/Python/3.5/lib/python/site-packages/neovim/api/common.py", line 138, in __getitem__
[deoplete]     return self._fetch()[idx]
[deoplete] IndexError: list index out of range
[deoplete] An error has occurred. Please execute :messages command.
```

tweekmonster:


rplugin/python3/deoplete/sources/buffer.py
```diff
@@ -39,22 +39,25 @@ def on_buffer(self, context):
 
     def __make_cache(self, context, bufnr):
         p = re.compile(context['keyword_patterns'])
+        buffer = self.vim.current.buffer
+        for b in reversed(self.vim.buffers):
+            if b.number < bufnr:
+                buffer = b
+                break
 
-        bufnr -= 1
```
tweekmonster: I think I misunderstood this.  Is the cache being created for the current buffer?

tweekmonster:
I think this PR isn't good because it's based on a misinterpretation of `RemoteSequence`.  `buffer.number` correlates with the actual buffer number in Vim, which isn't guaranteed to be contiguous.  `RemoteSequence` is 0 indexed, which I think was the original reason for `bufnr -= 1`.

It looks like iterating over `self.vim.buffers` will produce unnecessary remote calls to find a matching buffer.  Is there a reason you didn't stick with using `self.vim.current.buffer`?

tweekmonster:
I changed it to use `self.vim.current.buffer` instead of getting the buffer from `self.vim.buffers`.  If this doesn't work with your original intention feel free to close the PR 🙂

Shougo:
LGTM

PR #256, opened 2016-04-17
fix syntax error in vim-plug example

beaugunderson:
the missing comma causes a syntax error that's hard to track down. :)


larrylv:
Looks like we found the same problem :) https://github.com/Shougo/deoplete.nvim/pull/255

Shougo:
Thanks.

PR #259, opened 2016-04-19
Fix #258 support multiple filetype

### LINKED ISSUE #258: Deoplete sources do not handle composed filetypes
# Problems summary

Composed filetypes (as in "python.django") are not handled by deoplete.
See : https://github.com/zchee/deoplete-jedi/issues/33

I was wondering if it is by design, or if it could be possible to modify deoplete to manage those cases:
https://github.com/Shougo/deoplete.nvim/blob/master/rplugin/python3/deoplete/deoplete.py#L105-L106

## Expected

deoplete handle python specfic sources in combined filetypes buffers.

## Environment Information

 * OS: Ubuntu 15.10
 * Neovim version: v0.1.4-174-gba9bdb3

tweekmonster:


rplugin/python3/deoplete/util.py
```diff
@@ -145,3 +145,26 @@ def strwidth(string):
...
     wc = unicodedata.east_asian_width(c)
     return 2 if wc == 'F' or wc == 'W' else 1
+
+
+def itersource(vim, context, sources):
+    filetypes = context['filetype'].split('.')
+    ignore_sources = set()
+    for ft in filetypes:
+        ignores = get_buffer_config(vim, ft, 'b:deoplete_ignore_sources',
+                                    'g:deoplete#ignore_sources', '{}')
```
tweekmonster: I got inconsistent results from `deoplete#util#get_buffer_config()`.  The `b:` variable returned the entire dict, but the `g:` variable returned the dict value (list).  Is this intentional?

Shougo: The `b:` variable must be the source names list.

Shougo:
I will fix it after it merged.

PR #262, opened 2016-04-20
Fix buffer and member source flicker

tweekmonster:
Excludes current completion string from results to prevent this flickering:

flickering


Shougo:
Nice.

PR #268, opened 2016-04-28
Fix 'is_skip' to give first priority to the 'disabled_syntaxes' value

zchee:
If set `disabled_syntaxes`, but `input_pattern` takes precedence over it.
I was fixing to give first priority to the `disabled_syntaxes` value.

e.g. ( `|` is cursor )
```vim
call deoplete#custom#set('go', 'disabled_syntaxes', ['Comment'])
```

```go
package main

import (
	"fmt"
)


func main() {
	// fmt.|
	fmt.Println("test for deoplete.nvim")
}
```

Please feel free to point out.


Shougo:
Nice.

PR #276, opened 2016-05-14
Fix regex bug when input_pattern contains '|'

zhou13:
This should close https://github.com/zchee/deoplete-clang/issues/28 .


Shougo:
Thank you.

PR #280, opened 2016-06-01
Ignore empty abbr in delimiter auto converter

tweekmonster:
For completion sources that return a dict with empty `abbr` values, the delimiter auto converter will match all delimiters to the empty string.

The problem:

```
[x for x in needles if s.rfind(x) == (len(s) - len(x))]
```

`s.rfind(x)` will return `-1` when `s` is empty.  `len(s)` will be `0` and `len(x)` will be `1`.  This will evaluate to `-1 == -1`, which causes a match for _all_ 1 character delimiters.  So, the first delimiter will be used.

I also rewrote `last_find()` to stop at the first match instead of testing all delimiters.  Instead of using `.rfind()` which will scan the entire string, use `s[-1:]` instead to match expected region for the delimiter.  This will give a very minor performance increase.


Shougo:
LGTM

PR #283, opened 2016-06-02
Ignore permission error in file source

tweekmonster:
I hope you don't mind, but I also made a couple tweaks as well:

- Sort only once
- Use a lowercase sort key


Shougo:
LGTM

PR #289, opened 2016-06-13
Fix buffer config

paretje:
Since https://github.com/Shougo/deoplete.nvim/commit/4d18f71eb1f64668deced1d1e16f145045897ab6, the default filetype of the user config was ignored.


Shougo:
Thanks.

PR #296, opened 2016-06-21
Refactor

Shougo:
It is the refactored(file renamed) version of deoplete.


wsdjeg:
the build test failed, and there is an unknown function deoplete#mapping#refresh

Shougo:
I have fixed the problem.

wsdjeg:
nice, thanks for quick fix.

PR #303, opened 2016-07-11
Optimized path parsing in file source

tweekmonster:
`__longest_path_that_exists` was discarding `[^\f]` matches even though they could be part of an actual file path.  It's still split, but all of the original text is retained when checking paths.  Otherwise files with multiple spaces, escaped characters `\ `, or other non-standard characters that do exist won't work.

`__substitute_path` was using Vim to parse the path, but Python is well equipped to handle it.  `os.path.expanduser()` will work correctly even if `$HOME` doesn't exist.  Also, the pattern was allowing paths like `.../` to move up one directory per dot.  It's changed to only allow a max of 2 dots since this may be unexpected behavior.  If this was expected, let me know and I'll restore it.

### DIFF
--- a/rplugin/python3/deoplete/source/file.py
+++ b/rplugin/python3/deoplete/source/file.py
@@ -57,31 +57,18 @@ class Source(Base):
                 ] + [{'word': x} for x in files]
 
     def __longest_path_that_exists(self, context, input_str):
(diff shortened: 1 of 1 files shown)


Shougo:
Thank you.

PR #304, opened 2016-07-11
Handle completion delay in Vim instead of in Python

tweekmonster:
This uses a `timer_start()` in VimL to delay sending the context via `rpcnotify()` instead of using `time.sleep()` in Python.

@yrjan noticed that the `time.sleep()` was accumulating and causing long delays after typing a lot.  The other problem is that even if it was corrected to not block subsequent calls, the context would be old by the time completion happens.

Using `time.sleep()` with `500ms` delay to exaggerate the issue:

ezgif-8705343

Using `timer_start()` also has a nice side effect.  If you set the delay to `50ms` it will resolve the flickering when pressing backspace.  Maybe it can be set by default?

With `50ms` delay:

delay

No delay:

no delay

### DIFF
--- a/autoload/deoplete/handler.vim
+++ b/autoload/deoplete/handler.vim
@@ -24,7 +24,25 @@ function! deoplete#handler#_init() abort "{{{
   call s:on_event('')
 endfunction"}}}
 
-function! s:completion_begin(event) abort "{{{
+function! s:completion_begin_delayed(timer) abort "{{{
+  unlet! s:completion_delay
+  call s:completion_begin(s:delayed_event, 1)
+endfunction"}}}
+
+function! s:completion_begin(event, ...) abort "{{{
+  if has('timers') && !a:0 && g:deoplete#auto_complete_delay > 0
+    if exists('s:completion_delay')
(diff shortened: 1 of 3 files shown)


tweekmonster:
It looks like `timer_start()` isn't compatible with older versions of Neovim.  I'll think of another solution that could be handled in Python.

Shougo:
I must fix your changes.  Please wait.

tweekmonster:
@Shougo I came up with another way to handle the delay using Python.  Are you still looking into this?  Is it alright for me to start a new PR?

tweekmonster:
Here's the alternate method using Python's `threading.Timer` instead of `timer_start()`: https://github.com/tweekmonster/deoplete.nvim/commit/768df9565ced9112d721f50a02fc8702f7779e30

It has the same effect shown in the GIFs.  If it looks alright, I can force push into this PR and rename it.

Shougo:
> Here's the alternate method using Python's threading.Timer instead of timer_start(): tweekmonster@7a640ed

I think it is better.

Shougo:
Thanks.  It is better.

PR #307, opened 2016-07-17
Use a timer thread to handle delayed completion

tweekmonster:
Originally #304 

@yrjan noticed that the `time.sleep()` was accumulating and causing long delays after typing a lot.  The other problem is that even if it was corrected to not block subsequent calls, the context would be old by the time completion happens.

Using `time.sleep()` with `500ms` delay to exaggerate the issue:

ezgif-8705343

Using `timer_start()` also has a nice side effect.  If you set the delay to `50ms` it will resolve the flickering when pressing backspace.  Maybe it can be set by default?

With `50ms` delay:

delay

No delay:

no delay

### DIFF
--- a/autoload/deoplete/handler.vim
+++ b/autoload/deoplete/handler.vim
@@ -11,8 +11,8 @@ function! deoplete#handler#_init() abort "{{{
     autocmd CompleteDone * call s:complete_done()
     autocmd InsertCharPre * call s:on_insert_char_pre()
 
-    autocmd TextChangedI * call s:completion_begin("TextChangedI")
-    autocmd InsertEnter * call s:completion_begin("InsertEnter")
+    autocmd TextChangedI * call s:completion_begin("TextChangedI", 0)
+    autocmd InsertEnter * call s:completion_begin("InsertEnter", 0)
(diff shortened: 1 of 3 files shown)


Shougo:
LGTM

Shougo:
@tweekmonster I cannot reproduce the delay backspace behavior  in your screenshot.

tweekmonster:
Did you set `let g:deoplete#auto_complete_delay = 150` ?

Shougo:
> With 50ms delay?

I have set it to 50ms.  Why 150?

tweekmonster:
Ah, I'm sorry I forgot I didn't update the description.  50ms was the minimum that I tested that gave the desired results.  I found that 150ms was more consistent.

tweekmonster:
I also think that this method has a slightly greater delay than the `timer_start()` method.  In this method, the context is still created and `rpcnotify()` is called.  With `timer_start()`, it could prevent the context creation/parsing and `rpcnotify()`.

Shougo:
Hm.  VimL version is better.  I think.
This is slow.
I will revert it.

PR #311, opened 2016-07-18
Update README.md

lowski:
Include information about cache limit size if tags suggestions don't appear.


Shougo:
I have updated FAQ

PR #316, opened 2016-07-23
Fix file source base for new buffers

tweekmonster:
Use `cwd` if the buffer doesn't have a filename when buffer path completion is enabled.

### DIFF
--- a/rplugin/python3/deoplete/source/file.py
+++ b/rplugin/python3/deoplete/source/file.py
@@ -66,7 +66,11 @@ class Source(Base):
     def __substitute_path(self, context, path):
         m = re.match(r'(\.{1,2})/+', path)
         if m:
-            base = context['bufname'] if self.__buffer_path else context['cwd']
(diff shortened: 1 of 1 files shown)


rplugin/python3/deoplete/source/file.py
```diff
@@ -66,7 +66,11 @@ def __longest_path_that_exists(self, context, input_str):
     def __substitute_path(self, context, path):
         m = re.match(r'(\.{1,2})/+', path)
         if m:
-            base = context['bufname'] if self.__buffer_path else context['cwd']
+            if self.__buffer_path and context['bufname']:
+                base = context['bufname']
+            else:
+                base = os.path.join(context['cwd'], 'x')
```
tweekmonster: Forgot to mention in the PR message that it also adds a path component to `cwd` so that `./` and `../` should work correctly when not using the buffer path option.

Shougo:
Thanks

PR #317, opened 2016-07-23
Fix comment detection

tweekmonster:
`deoplete#util#get_syn_name()` was renamed to `deoplete#util#get_syn_names()`


Shougo:
Thanks

PR #326, opened 2016-08-07
Source refactor

tweekmonster:
- Refactored the source a bit to move some logic into `util` functions.
- Added `deoplete.util.error_tb()` that prints the stack trace before the message.
- Added exception handling in the plugin loading and completion gathering.  If they fail, they're removed instead of causing all completions to fail.  This will also make source development more bearable.
- Refactored `set_source_attributes()` so it'll be easier to manage the source attributes.
- When a source/filter is loaded, its path is added to `sys.path`.  This should remove the need for `load_external_module()` in _most_ cases.  Simple sources shouldn't have to worry about `sys.path`.

### DIFF
--- a/rplugin/python3/deoplete/deoplete.py
+++ b/rplugin/python3/deoplete/deoplete.py
@@ -5,23 +5,20 @@
 # ============================================================================
 
 from deoplete.util import \
-    error, globruntime, charpos2bytepos, \
+    error, error_tb, find_rplugins, import_plugin, charpos2bytepos, \
     bytepos2charpos, get_custom, get_syn_names, get_buffer_config
 
-import deoplete.source
-import deoplete.filter
-import deoplete.util
+import deoplete.source  # noqa
+import deoplete.filter  # noqa
+import deoplete.util  # noqa
 from deoplete import logger
 
 import re
-import importlib.machinery
 import os.path
 import copy
-import traceback
 import time
 
-deoplete.source  # silence pyflakes
-deoplete.filter  # silence pyflakes
+from collections import defaultdict
 
 
 class Deoplete(logger.LoggingMixin):
@@ -34,16 +31,15 @@ class Deoplete(logger.LoggingMixin):
         self.__custom = []
         self.__profile_flag = None
         self.__profile_start = 0
+        self.__source_errors = defaultdict(int)
+        self.__filter_errors = defaultdict(int)
         self.name = 'core'
 
     def completion_begin(self, context):
         try:
             complete_position, candidates = self.gather_candidates(context)
         except Exception:
-            for line in traceback.format_exc().splitlines():
-                error(self.__vim, line)
-            error(self.__vim,
-                  'An error has occurred. Please execute :messages command.')
+            error_tb(self.__vim, 'Error while gathering completions')
             candidates = []
 
         if not candidates or self.position_has_changed(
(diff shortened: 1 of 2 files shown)


rplugin/python3/deoplete/deoplete.py
```diff
@@ -68,96 +64,109 @@ def gather_candidates(self, context):
...
+                # Filtering
+                ignorecase = ctx['ignorecase']
+                smartcase = ctx['smartcase']
+                camelcase = ctx['camelcase']
+
                 # Set ignorecase
-                if (smartcase or camelcase) and re.search(
-                        r'[A-Z]', context['complete_str']):
-                    context['ignorecase'] = 0
+                if (smartcase or camelcase) \
```
Shougo: I don't like to add `\`...

rplugin/python3/deoplete/deoplete.py
```diff
@@ -235,108 +244,100 @@ def profile_end(self, name):
...
-                self.__filters[filter.name] = filter
-                self.debug('Loaded Filter: %s (%s)',
-                           filter.name, module.__file__)
-        # self.debug(self.__filters)
+        for path in find_rplugins(context, 'filter'):
+            name = os.path.splitext(os.path.basename(path))[0]
+            if name in self.__filters:
+                continue
+
+            filter_ = None
```
Shougo: `filter_` is the bad name.  Please use another name.

rplugin/python3/deoplete/util.py
```diff
@@ -52,6 +55,43 @@ def globruntime(runtimepath, path):
     return ret
 
 
+def find_rplugins(context, source):
+    """Search for base.py or *.py
+
+    Searches $VIMRUNTIME/*/rplugin/python3/deoplete/$source[s]/
+    """
+    rtp = context.get('runtimepath', '').split(',')
+    if rtp:
```
Shougo: It should be:

```python
if not rtp:
   return

...
```

rplugin/python3/deoplete/deoplete.py
tweekmonster: Okay I'll change it.

The reason I do it is because I don't like the alternative:

```python
if (smartcase or camelcase) and re.search(r'[A-Z]',
                                          ctx['complete_str']):
    ctx['ignorecase'] = 0
```

Without braces, I think it makes it harder to read because the lines appear to be separated by whitespace.  The whitespace separation increases if the far right function has a lot of arguments.

rplugin/python3/deoplete/deoplete.py
tweekmonster: Ah, I thought I changed that before I committed.  It's something that's done in Python's standard library when a name conflicts with a builtin/keyword.  Like `class_`.

I'll change it.

This reminds me of another problem when I was trying to look for a simple solution to #327.  The use of the `__` for the "private" variables is considered to be internal magic like `__str__()` or `__eq__()`.  It doesn't need to be changed, but it made it impossible for me to check `hasattr(self, '__vim')` to figure out the logging issue.

rplugin/python3/deoplete/util.py
tweekmonster: 👍  This is a habit I have because the function is a generator, but it's only a problem if it returns a value.  I'll change it.

tweekmonster:
Updated

Shougo:
LGTM.  Thanks.

PR #329, opened 2016-08-13
Use paths that are more "environment-friendly"

andsild:
Regarding commit 39a89f4: 
http://stackoverflow.com/a/16365367
http://stackoverflow.com/a/21613044

Commit 0282b16:
/usr/lib doesn't exist on my system so the test fails (NixOS). /usr/bin has to exist on all UNIX systems

### DIFF
--- a/run_tests.sh
+++ b/run_tests.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
 set -e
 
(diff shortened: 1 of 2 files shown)


run_tests.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
```
Shougo: Why did you change it?
I don' think this change is needed.

tweekmonster: I've run into cases where `/bin/bash` doesn't actually exist and `/usr/bin/env bash` was more reliable.  I think it was on FreeBSD.

OTH, the contents of `run_tests.sh` could be placed in `Makefile`.

Shougo:
Thank you.  I have fixed.

PR #332, opened 2016-08-19
use pkg_resources to find neovim client version on startup

tek:
the method used so far fails when using a local checkout of neovim-client


Shougo:
LGTM

PR #343, opened 2016-09-24
Fix Neovim version check.

tweekmonster:
fixes #342

PR #344, opened 2016-09-24
Fix neovim-python version check

imipolexg:
Deoplete.nvim was using a lexical comparison of the neovim-python
version which caused it to report version 0.1.10 as older than 0.1.8.
This replaces it with a numerical comparison.


imipolexg:
Just noticed https://github.com/Shougo/deoplete.nvim/pull/343, so closing this.

PR #345, opened 2016-09-28
Ensure that b:changedtick is unchanged with delayed completion

blueyed:
This adds a short-circuit for the b:changedtick check, but also uses the
tick from when the timer was started.

This is important with manual triggering of completions using e.g.
`<C-p>`, and can be seen best with long timeouts (i.e. `:let
g:deoplete#auto_complete_delay = 1000`): opening of the completion popup
triggers b:changedtick and this prevents deoplete from stomping over the
existing completions.

I have also moved s:completion_delayed down to make it easier to follow
the code.

Apart from the issue this fixes I think that having a short-circuit for
the common case of checking b:changedtick is good to have - skipping the
building of the context all the time.

### DIFF
--- a/autoload/deoplete/handler.vim
+++ b/autoload/deoplete/handler.vim
@@ -24,12 +24,6 @@ function! deoplete#handler#_init() abort "{{{
   call s:on_event('')
 endfunction"}}}
 
-function! s:completion_delayed(timer) abort "{{{
-  let timer = s:timer
-  unlet! s:timer
-  call s:completion_begin(timer.event)
-endfunction"}}}
-
 function! s:completion_check(event) abort "{{{
   let delay = get(g:deoplete#_context, 'refresh', 0) ?
         \ g:deoplete#auto_refresh_delay : g:deoplete#auto_complete_delay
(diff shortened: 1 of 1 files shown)


autoload/deoplete/handler.vim
```diff
@@ -48,7 +42,16 @@ function! s:completion_check(event) abort "{{{
   return s:completion_begin(a:event)
 endfunction"}}}
 
-function! s:completion_begin(event) abort "{{{
+function! s:completion_delayed(timer) abort "{{{
+  let timer = s:timer
+  unlet! s:timer
+  call s:completion_begin(timer.event, timer.changedtick)
```
Shougo: It should be

```vim
if b:changedtick != timer.changedtick
  call s:completion_begin(timer.event, timer.changedtick)
endif
```

autoload/deoplete/handler.vim
```diff
@@ -48,7 +42,16 @@ function! s:completion_check(event) abort "{{{
...
+endfunction"}}}
+
+function! s:completion_begin(event, ...) abort "{{{
+  if a:0 && b:changedtick != a:1
```
Shougo: Please remove the block.
It should be checked in `s:completion_delayed()`.

blueyed:
@Shougo 
Updated.

Shougo:
LGTM

PR #349, opened 2016-10-08
Fix quotes around still_have_issues health check message.

brcolow:
The error message is currently rendering incorrectly as you can see here:

test

I *think* I did this correctly - but I didn't test it. However, I wanted to open a PR instead of just posting an issue. Please review it carefully :).


Shougo:
LGTM

PR #351, opened 2016-10-12
Simplify condition for omnifunc check

thalesmello:
I was reviewing the source code, and found this condition that could be simplified.


Shougo:
LGTM

PR #354, opened 2016-10-13
Try file path before non-word characters.

tweekmonster:
This allows file matches on text like `set rtp+=~/`


wsdjeg:
I can complete files after `set rtp+=~/` without this patch,can you try add `set isfname-==
` in your vimrc?

tweekmonster:
@wsdjeg I didn't want to modify `isfname` since it is possible for `+` and `=` to be in a path.  The files source only uses `isfname` as a hint.  It's different from `^X^F` so I don't think modifying `isfname` should be necessary.

Shougo:
OK.

PR #360, opened 2016-10-20
Add abort option to buffer source

dalance:
When I edit a very large file ( ex. 2GB, 200000000line ), `__make_cache` of buffer.py consumes long time, and finally the following error occurs.

```
Error detected while processing function <SNR>96_completion_delayed[4]..<SNR>96_completion_begin:
E475: Invalid argument: Channel doesn't exist
```

I don't need buffer source at such large files, so I add the threshold of buffer length for aborting `__make_cache`.

### DIFF
--- a/rplugin/python3/deoplete/source/buffer.py
+++ b/rplugin/python3/deoplete/source/buffer.py
@@ -20,6 +20,7 @@ class Source(Base):
         self.mark = '[B]'
         self.__buffers = {}
         self.__max_lines = 5000
+        self.__abort_lines = 1000000
 
     def on_event(self, context):
         if ((context['bufnr'] not in self.__buffers) or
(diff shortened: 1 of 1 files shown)


tweekmonster:
You should be able to use this to deal with this:

```vim
autocmd BufRead * if getline('$') > 1000000 | let b:deoplete_ignore_sources = ['buffer'] | endif
```

Just a thought: Isn't a million lines still excessive?  Also, wouldn't loading a 2GB file in nvim be a bigger issue than the script host crashing?  I think it's good to guard against things that's certain to crash, but I feel like this is an edge case.  I typically get impatient and kill the process after I accidentally open anything over 1MB that causes a 5 second pause since it most likely means doing anything after it loads is going to suck (syntax, autoident, foldexpr, etc).

In any case, that should probably be a user-configurable setting since lower values might be better on slower systems.

dalance:
@tweekmonster 
I understand your solution. Thank you.

Issue #346, opened 2016-09-29
Undo breaks after completion


tweekmonster:
It seems `<Plug>(deoplete_start_complete)` breaks undo.

## Environment Information

* OS: Ubuntu 16.04.1 LTS
* Nvim: v0.1.6-167-g22dfe69

## Reproduction

- Use standard minimal init.
- Trigger completion (.e.g `ca` in a Vim script)
- When the completion menu is displayed, press `<esc>`
- Press `A` in normal mode to re-enter insert mode and trigger completion

For every time `<esc>A` is pressed, a new undo is created.

## Solution?

If I add `undojoin` above [line 17](https://github.com/Shougo/deoplete.nvim/blob/d53c8f46f1aff5e015076c86dbcdfcc35d5a9015/autoload/deoplete/mapping.vim#L17) in `mapping.vim`, the issue is resolved.  But, I'm not sure if there are cases where `undojoin` will cause an error.

```diff
diff --git a/autoload/deoplete/mapping.vim b/autoload/deoplete/mapping.vim
index a14dd1d..8c4cc1d 100644
--- a/autoload/deoplete/mapping.vim
+++ b/autoload/deoplete/mapping.vim
@@ -14,6 +14,7 @@ endfunction"}}}
 function! deoplete#mapping#_do_complete(context) abort "{{{
   if b:changedtick == get(a:context, 'changedtick', -1)
         \ && mode() ==# 'i'
+    undojoin
     call complete(a:context.complete_position + 1, a:context.candidates)
   endif
```

tweekmonster:
[Here's a video of the problem.](https://share.esdf.io/lvjoYa9p0u/Screen-Recording-2016-09-29-10-28-37.mp4)

tweekmonster:
@Shougo Sorry I just noticed a problem.  I think this needs a revert.  When I complete with a `.` after undo, the undo state cause `undojoin` to generate an error.  I'm not sure what the difference is, but maybe it's because `.` doesn't wait for a delay?

Maybe the context needs to include a variable to tell the `_do_complete()` function that it was a delayed trigger from the rplugin?

Shougo:
> Maybe the context needs to include a variable to tell the _do_complete() function that it was a delayed trigger from the rplugin?

Please test the change.
I have not time to test it.

tweekmonster:
[...]

PR #362, opened 2016-10-24
Remove undojoin

tweekmonster:
This issue was resolved in https://github.com/neovim/neovim/pull/5499

Original issue: #346 

The undo will remain inconsistent for nvim < 0.1.6, but it seems like I was the only person who noticed/cared.


Shougo:
OK

PR #369, opened 2016-11-02
buffer: let user disable same-filetype restriction

sunaku:
This patch allows the user to disable the buffer source's same-filetype
restriction, which limits keyword collection to same-filetype buffers by
default, by exposing a new `require_same_filetype` custom setting:

    " Collect keywords from all buffers, regardless of filetype.
    call deoplete#custom#set('buffer', 'require_same_filetype', 0)

Without this patch, to complete a word from a different filetype buffer,
the user needs to switch away from Deoplete to Vim's native completion:

1. Wait for Deoplete to present its completion menu
2. Press `<C-E>` to abort Deoplete's completion menu
3. Press `<C-N>` to activate Vim's native completion

### DIFF
--- a/doc/deoplete.txt
+++ b/doc/deoplete.txt
@@ -519,8 +519,8 @@ buffer						*deoplete-source-buffer*
 		It collects keywords from current buffer and buffers which
 		have same 'filetype', unless you specify this custom setting:
 >
-		" Collect keywords from all buffers, regardless of filetype.
-		call deoplete#custom#set('buffer', 'require_same_filetype', 0)
+		" Collect keywords from buffers of any filetype.
+		let g:deoplete#buffer_require_same_filetype = 0
 <
 		Note: It takes time to get the candidates in the first time if
 		you want to edit the large files(like Vim 22000 lines eval.c).
--- a/doc/deoplete.txt
+++ b/doc/deoplete.txt
@@ -516,8 +516,12 @@ EXAMPLES						*deoplete-examples*
 SOURCES							*deoplete-sources*
 
 buffer						*deoplete-source-buffer*
-		It collects keywords from current buffer and the buffers which
-		have same 'filetype'.
+		It collects keywords from current buffer and buffers which
+		have same 'filetype', unless you specify this custom setting:
+>
+		" Collect keywords from all buffers, regardless of filetype.
+		call deoplete#custom#set('buffer', 'require_same_filetype', 0)
+<
 		Note: It takes time to get the candidates in the first time if
 		you want to edit the large files(like Vim 22000 lines eval.c).
 
--- a/rplugin/python3/deoplete/source/buffer.py
+++ b/rplugin/python3/deoplete/source/buffer.py
@@ -29,8 +29,12 @@ class Source(Base):
     def gather_candidates(self, context):
         self.__make_cache(context)
 
-        buffers = [x['candidates'] for x in self.__buffers.values()
-                   if x['filetype'] in context['filetypes']]
+        buffer_values = self.__buffers.values()
+        if context['vars'].get('require_same_filetype', True):
+          buffer_values = [x for x in buffer_values
+                           if x['filetype'] in context['filetypes']]
+
+        buffers = [x['candidates'] for x in buffer_values]
         if not buffers:
             return []
 


rplugin/python3/deoplete/source/buffer.py
```diff
@@ -29,8 +29,12 @@ def on_event(self, context):
     def gather_candidates(self, context):
         self.__make_cache(context)
 
-        buffers = [x['candidates'] for x in self.__buffers.values()
-                   if x['filetype'] in context['filetypes']]
+        buffer_values = self.__buffers.values()
+        if context['vars'].get('require_same_filetype', True):
+          buffer_values = [x for x in buffer_values
```
Shougo: Wronged indentation.

doc/deoplete.txt
```diff
@@ -516,8 +516,12 @@ EXAMPLES						*deoplete-examples*
 SOURCES							*deoplete-sources*
 
 buffer						*deoplete-source-buffer*
-		It collects keywords from current buffer and the buffers which
-		have same 'filetype'.
+		It collects keywords from current buffer and buffers which
+		have same 'filetype', unless you specify this custom setting:
+>
+		" Collect keywords from all buffers, regardless of filetype.
+		call deoplete#custom#set('buffer', 'require_same_filetype', 0)
```
Shougo: It does not work.
Have you tested it?

It should be `g:deoplete#buffer_require_same_filetype` variable.

Shougo:
It seems good feature.
But it has the wrong implementation.

sunaku:
Thanks for your feedback! :+1: I will fix these problems and update this PR accordingly. :sweat_smile:

rplugin/python3/deoplete/source/buffer.py
tweekmonster: @sunaku This isn't entirely due to your changes, but this can be optimized like so:

```python
import itertools

...

        same_filetype = context['vars'].get('require_same_filetype', False)
        candidates = (x['candidates'] for x in self.__buffers.values()
                      if not same_filetype or
                      (same_filetype and x['filetype'] in context['filetypes']))
        return [{'word': x} for x in itertools.chain(*candidates)]
```

The old code + your changes has two list comprehensions (e.g. `[x for x in y]`).  With, the `functools.reduce()` line, you end up with 3 list iterations as well as `reduce()` creating new lists on each `add()`.

Using a generator comprehension (e.g. `(x for x in y)`), the results are evaluated when expanding `*candidates` on `itertools.chain()`.  In addition, the candidate lists are iterated and yielded to the final list comprehension (as opposed to concatenated *then* iterated).  Also, `functools.reduce()` throws an error on an empty list, `itertools.chain()` will just yield nothing, so the check for an empty list can be skipped.

I'm not generally uptight about optimizations like this, but I think it's worth shaving off a few nano/milliseconds for live completions.  You don't have add this code now if you don't want to.  But, I'll most likely add it myself in a separate PR if you don't 😜

tweekmonster: Sorry, in the snippet I posted above, I set the default for `require_same_filetype` to `False`, but I think it makes more sense if it's `True` by default.

sunaku:
@Shougo you are correct: this patch doesn't work. :cold_sweat:  Even forcibly disabling the filetype restriction (as follows) has no visible effect on Deoplete's completion menu: it simply doesn't search keywords in nearby buffers, regardless of same or different filetype:

```diff
diff --git a/rplugin/python3/deoplete/source/buffer.py b/rplugin/python3/deoplete/source/buffer.py
index 3ed4eb8..7f77d10 100644
--- a/rplugin/python3/deoplete/source/buffer.py
+++ b/rplugin/python3/deoplete/source/buffer.py
@@ -29,8 +29,7 @@ class Source(Base):
     def gather_candidates(self, context):
         self.__make_cache(context)

-        buffers = [x['candidates'] for x in self.__buffers.values()
-                   if x['filetype'] in context['filetypes']]
+        buffers = [x['candidates'] for x in self.__buffers.values()]
         if not buffers:
             return []
```

So the key to implementing this feature _is not the above_ location, as I hoped. :crying_cat_face:  Instead, it's  somewhere else in the source code. :earth_asia:  This needs further exploration. :timer_clock:

rplugin/python3/deoplete/source/buffer.py
tweekmonster: Oops, I think it can just be:

```python
if not same_filetype or x['filetype'] in context['filetypes']
```
instead of

```python
if not same_filetype or (same_filetype and x['filetype'] in context['filetypes'])
```

Right?  I think I need to get some sleep.

sunaku: Righteous tweaks! :+1:  Go ahead with your PR because this patch doesn't seem to cut it anyway. :8ball:

tweekmonster:
@sunaku Are you *saving* the nearby buffers before testing?  Look at the `on_event()` function.  It doesn't update the cache until a buffer is saved.  I'm getting completions from `vim` buffers in `python` buffers.

sunaku:
Aha! Thanks @tweekmonster, that did the trick! :relieved:

This is rather unexpected; I thought it would search buffers on demand, like native Vim completion does. :sweat:  Time for another PR I guess.

Shougo:
@tweekmonster Thank you.  I have imported the changes.

@sunaku It must be

```python
same_filetype = context['vars'].get('deoplete#buffer_require_same_filetype', True)
```

doc/deoplete.txt
```diff
@@ -516,8 +516,12 @@ EXAMPLES						*deoplete-examples*
 SOURCES							*deoplete-sources*
 
 buffer						*deoplete-source-buffer*
-		It collects keywords from current buffer and the buffers which
-		have same 'filetype'.
+		It collects keywords from current buffer and buffers which
+		have same 'filetype', unless you specify this custom setting:
+>
+		" Collect keywords from buffers of any filetype.
+		let g:deoplete#buffer_require_same_filetype = 0
```
Shougo: let g:deoplete#buffer#require_same_filetype = 0

Shougo:
I have implemented it.

PR #383, opened 2016-11-18
Make comparison operators user-setting insensitive

Alok:
Changed all operators to strict variants, so user's `ignorecase`
settings don't have to be taken into account.


Shougo:
I don't want to add `!=# ''`.
It is my style.

PR #384, opened 2016-11-19
Do not call deoplete#init#_enable twice with

blueyed:
With g:deoplete#enable_at_startup, `deoplete#init#_enable` gets called
twice, which results in the sources being initialized twice:

```
deoplete#enable
deoplete#initialize
deoplete#init#_initialize
deoplete#init#_enable
deoplete#handler#_init
s:on_event ''
deoplete#init#_enable
deoplete#handler#_init
s:on_event ''
```


Shougo:
OK

PR #392, opened 2016-11-29
Improve error handling with on_init/on_event

blueyed:
It will ignore sources that have thrown an exception in `on_init`, and
provides an error message in that case.

With `on_event` the error gets displayed.


Shougo:
Merged

PR #394, opened 2016-12-01
minor: vint fixes

blueyed:


Shougo:
I don't want to add `==# ''` compare.

PR #401, opened 2016-12-18
Ignore sources that have None for the filetypes attribute

tweekmonster:
Ignore any source that sets `filetypes` to `None`.  If it's `None` the source should be considered broken or not correctly configured.  Sources that work in all files shouldn't set `filetypes` since it's the default behavior.

Related: https://github.com/carlitux/deoplete-ternjs/pull/37


Shougo:
OK.

PR #410, opened 2017-01-04
No changes are made to completeopt if there isn't enough characters for completion

purpleP:
@Shougo 
I've hardcoded number `2` there, but this of course should be a setting. I think you even already have one called `min_pattern_length`, but I'm not sure what this settings means, so I've hardcoded 2 and you can change that to whatever setting name you want.

I've tested it manually. Works fine for me.

### DIFF
--- a/doc/deoplete.txt
+++ b/doc/deoplete.txt
@@ -533,8 +533,9 @@ EXAMPLES						*deoplete-examples*
 SOURCES							*deoplete-sources*
 
 buffer						*deoplete-source-buffer*
-		It collects keywords from current buffer and the opened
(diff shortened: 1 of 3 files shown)


Shougo:
Please fix lint error.
Your code seems broken.

I will test your branch.

> I've hardcoded number 2 there, but this of course should be a setting. I think you even already have one called min_pattern_length, but I'm not sure what this settings means, so I've hardcoded 2 and you can change that to whatever setting name you want.

It must be fixed.  OK.  I will check the code.

purpleP:
@Shougo I think I fixed linter error. While hacking I've just forgot about unused function

PR #411, opened 2017-01-05
Feature/completeopt save

purpleP:
This should fix linter erros. Somehow vint isn't working for me on python3.5, so I can't be sure.


Shougo:
I have implemented the improved version of the feature.

purpleP:
@Shougo Great, thanks!

PR #417, opened 2017-01-14
Add and handle SourceInitError

blueyed:
This is meant to be by sources (e.g. deoplete-jedi) in case of clear
errors.

This also adds the formatting of `exc` to the fallback's / previously
used `error_tb`.


Shougo:
> This also adds the formatting of exc to the fallback's / previously
used error_tb.

Why it uses `error()` instead of `error_tb()`?
I don't understand it.

blueyed:
> Why it uses error() instead of error_tb()?

To make it less intrusive: the error is clear, and therefore there should not be an additional traceback.

Shougo:
OK.  Merged.

PR #438, opened 2017-02-23
improve doc

e-kwsm:
* s/Number/Integer/  
"Number" seems to include float; at least the lengths and the widths should be int
* s/Unknown/Unspecified/


e-kwsm:
Just for information: annotating types in [typing](https://docs.python.org/3/library/typing.html) way, e.g. `List[str]`, may make easier to create sources.

Shougo:
Merged.

> Just for information: annotating types in typing way, e.g. List[str], may make easier to create sources.

OK.

PR #442, opened 2017-03-04
Support subdirectories for source and filter

e-kwsm:
vim allows subdirectories under `after`, e.g. `after/syntax/c/{one,two}.vim`; this PR is analogy of that.
Of course we can separate files like `rplugin/python3/deoplete/source/c_{one,two}.py`, but it is convenient if subdirectories are available.

### DIFF
--- a/rplugin/python3/deoplete/util.py
+++ b/rplugin/python3/deoplete/util.py
@@ -71,6 +71,7 @@ def find_rplugins(context, source):
         os.path.join('rplugin/python3/deoplete', source, 'base.py'),
         os.path.join('rplugin/python3/deoplete', source, '*.py'),
         os.path.join('rplugin/python3/deoplete', source + 's', '*.py'),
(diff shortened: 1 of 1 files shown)


rplugin/python3/deoplete/util.py
```diff
@@ -71,6 +71,8 @@ def find_rplugins(context, source):
         os.path.join('rplugin/python3/deoplete', source, 'base.py'),
         os.path.join('rplugin/python3/deoplete', source, '*.py'),
         os.path.join('rplugin/python3/deoplete', source + 's', '*.py'),
+        os.path.join('rplugin/python3/deoplete', source, '*', '*.py'),
+        os.path.join('rplugin/python3/deoplete', source + 's', '*', '*.py'),
```
Shougo: Please remove the line.
Because, `sources` are just for the compatibility.

e-kwsm:
Is this OK?

Shougo:
Merged.

PR #445, opened 2017-03-11
matcher_length should include same size matches

docwhat:
`matcher_length` was filtering out candidates that were the same size as
the completion string.

For example, the string "defr" wasn't returning the exact match while "def"
would return "defr" as a match.

### DIFF
--- a/rplugin/python3/deoplete/filter/matcher_length.py
+++ b/rplugin/python3/deoplete/filter/matcher_length.py
@@ -18,4 +18,4 @@ class Filter(Base):
     def filter(self, context):
         input_len = len(context['complete_str'])
         return [x for x in context['candidates']
-                if len(x['word']) > input_len]
+                if len(x['word']) >= input_len]


Shougo:
Why this feature is needed?  I don't understand.
You can disable `matcher_length`.

docwhat:
I'm sorry.  I didn't explain it.

The documentation says:

> It removes shorter candidates than user input.

But what it was actually doing was removing candidate equal to or shorter than the user input.

This commit fixes the code to match the documentation.

I feel the documentation version is the more correct behavior, but you may feel differently.

roxma:
IMO keeping candidates equal to user input should be useful sometimes, as a hint that my typing is right.

docwhat:
I agree.  But the current code doesn't do that.

If my input is 3 characters then only candidates of 4+ are shown.

My change will show candidates of 3+

docwhat:
Documentation fixed in 306bf5bf4583faefe1a1a788d643e9a5f30743e6

Shougo:
```
Q: I want to see the typed word in the completion menu.

A: You should remove |deoplete-filter-matcher_length| from the matchers.

	call deoplete#custom#set('_', 'matchers', ['matcher_fuzzy'])

```

docwhat:
I missed that FAQ item.  Sorry to waste your time.

Meanwhile, I created a plugin that does what I want, in case others want it too: https://github.com/docwhat/deoplete-better-length-matcher.nvim

wsdjeg:
I agree with shougo, why we need completion plugin, it help us complete code, it is not lint, if I REMEMBER what the code is, why should the popmenu should be opened? also if I REMEMBER the code and I am fast typing, the popmenu also should not be opened, that is why some people think deoplete is slow, and that is why I do not use nvim complete manager.

docwhat:
In my case because I wasn't sure of the snippet I was filling in.  I agree, it is useless for normal completion (I completed it already) but harmless since if I remember it and am typing past it, deoplete usually doesn't have time to do a popup.

Hmm... maybe I can make my version be my rule for snippets and the original rule for non-snippets?

Shougo:
> the popmenu also should not be opened, that is why some people think deoplete is slow, and that is why I do not use nvim complete manager.

I have improved deoplete completion speed :-)

https://github.com/Shougo/deoplete.nvim/pull/447

Shougo:
> In my case because I wasn't sure of the snippet I was filling in. I agree, it is useless for normal completion (I completed it already) but harmless since if I remember it and am typing past it, deoplete usually doesn't have time to do a popup.

You can disable `matcher_length` in snippets source only.
So, this is no problem.

docwhat:
D'oh!

wsdjeg:
@roxma
why do you block me on github? just because i said mcwm do not has delay feature? I just want to tell you what feature  I need with your plugin.
ok, hope ncwm be better!

PR #447, opened 2017-03-12
Improve the refresh behavior

### LINKED ISSUE #444: Ultisnips <c-j> tabstop not working for snippet inside snippet
# Problems summary

Before describing the problem, I would like to save your time telling you the temporary solution I found. 
Write the following line in your init.vim and the problem halts:
```
let g:deoplete#auto_complete_delay=150 " or something higher than 100
```
Yeap. As you might have noticed, I think this is somehow related to #441 and #440 implicitly. It is to be noted that the above code only works as a partial solution; I still get the symptoms. It just makes it way less frequent.

[I posted the problem thoroughly in this link.](https://github.com/SirVer/ultisnips/issues/831)

For those who hate to click links, here is a partial excerpt:
---
Please consider these simple latex snippets:
```
snippet sqr "sqrt"
\sqrt{$1}$0
endsnippet

snippet alp "alpha" i
\alpha
endsnippet
```

If I open a new tex file named "test.tex" and type the following:
snip -----
`sqr<tab>alp<tab><c-j>`
snap ------
```
\sqrt{\alpha}
             ^ (cursor position)
```

It works as expected and I'm satisfied. 
### But the problem arises when I try the same snippet again somewhere else in the same file.
snip ----- (identical)
`sqr<tab>alp<tab><c-j>`
snap -----
```
\sqrt{\alpha}
      ^    ^ (cursor in SELECT mode)
   from    to
```
The `<c-j>` (jump forward) key behaves strange. Instead of jumping to the defined $0 of sqr snippet, it visually selects `\alpha` and turns VIM into SELECT mode. This happens over and over if the same snippet is triggered.
Even if I restart vim (= quit the file with `:wq` and come back to this file with `vim test.tex`), this odd behavior still remains. The only way that I can restore back to its first, normal, originally intended behavior is to delete all the related snippets (or delete all lines using `ggdG`) and restart vim. Then the first attempt works well, while the second trial repeats the odd behavior, again and again.

**It is to be noted that this is only a partial excerpt of the link. More information available in the link, if needed.**

---

## Expected

snip -----
`sqr<tab>alp<tab><c-j>`
snap ------
```
\sqrt{\alpha}
             ^ (cursor position)
```
No matter how many times sqr and alp are triggered, their tabstops must be functioning normal.

## Environment Information

 * deoplete version(SHA1):

 * OS: macOS

 * neovim version: 0.1.7

 * `:CheckHealth` result(neovim ver.0.1.5-452+):
```

health#deoplete#check
========================================================================
## deoplete.nvim
  - SUCCESS: has("nvim") was successful
  - SUCCESS: has("python3") was successful
  - INFO: If you're still having problems, try the following commands:
    $ export NVIM_PYTHON_LOG_FILE=/tmp/log
    $ export NVIM_PYTHON_LOG_LEVEL=DEBUG
    $ nvim
    $ cat /tmp/log_{PID}
    and then create an issue on github

health#nvim#check
========================================================================
## Configuration
  - SUCCESS: no issues found

## Performance
  - SUCCESS: Build type: RelWithDebInfo

## Remote Plugins
  - SUCCESS: Up to date

## terminfo
  - ERROR: key_backspace (kbs) entry is ^H (ASCII DELETE): key_backspace=^H,
    - SUGGESTIONS:
      - Set key_backspace to \177 (ASCII BACKSPACE). Run these commands:
          infocmp $TERM | sed 's/kbs=^[hH]/kbs=\\177/' > $TERM.ti
          tic $TERM.ti
      - See https://github.com/neovim/neovim/wiki/FAQ

## tmux
  - SUCCESS: escape-time: 0ms
  - INFO: $TERM: screen-256color

health#provider#check
========================================================================
## Clipboard
  - SUCCESS: Clipboard tool found: pbcopy

## Python 2 provider
  - INFO: `g:python_host_prog` is not set.  Searching for python2 in the environment.
  - INFO: Executable: /usr/local/bin/python2
  - INFO: Python2 version: 2.7.13
  - INFO: python2-neovim version: 0.1.13
  - SUCCESS: Latest python2-neovim is installed: 0.1.13

## Python 3 provider
  - INFO: `g:python3_host_prog` is not set.  Searching for python3 in the environment.
  - INFO: There are multiple python3 executables found.  Set "g:python3_host_prog" to avoid surprises.
  - INFO: Executable: /usr/local/bin/python3
  - INFO: Other python executable: /Library/Frameworks/Python.framework/Versions/3.5/bin/python3
  - INFO: Python3 version: 3.5.2
  - INFO: python3-neovim version: 0.1.13
  - SUCCESS: Latest python3-neovim is installed: 0.1.13

## Ruby provider
  - ERROR: Missing Neovim RubyGem
    - SUGGESTIONS:
      - Install or upgrade the neovim RubyGem using `gem install neovim`.
  - INFO: Ruby Version: ruby 2.4.0p0 (2016-12-24 revision 57164) [x86_64-darwin15]
  - INFO: Host Executable: not found
  - INFO: Host Version: not found


```

## Provide a minimal init.vim with less than 50 lines (Required!)

Partly described in the link.  
I use Vimplug.
If I comment out `Plug 'Shougo/deoplete.nvim' { 'do': ':UpdateRemotePlugins'}` everything works fine,
while if I uncomment the line, the problem pops out again.
I thought that will do.
If you really need my init.vim I will let you know.
## The reproduce ways from neovim starting (Required!)
```
$ vim newfile1.tex 
insert mode
sqr<tab>alp<tab><c-j>
repeat twice
the first one works
while the second doesn't
```

Shougo:
Fix ~#444~ and #440.
Please test it.

Note:  It improves the deoplete performance.

### DIFF
--- a/autoload/deoplete.vim
+++ b/autoload/deoplete.vim
@@ -44,26 +44,16 @@ function! deoplete#manual_complete(...) abort
         \ . string(get(a:000, 0, [])) . ")\<CR>"
 endfunction
 function! deoplete#close_popup() abort
-  if exists('g:deoplete#_context')
-    let g:deoplete#_context.position = getpos('.')
-  endif
   return pumvisible() ? "\<C-y>" : ''
 endfunction
 function! deoplete#smart_close_popup() abort
-  if exists('g:deoplete#_context')
-    let g:deoplete#_context.position = getpos('.')
-  endif
   return pumvisible() ? "\<C-e>" : ''
 endfunction
 function! deoplete#cancel_popup() abort
-  if exists('g:deoplete#_context')
-    let g:deoplete#_context.position = getpos('.')
-  endif
   return pumvisible() ? "\<C-e>" : ''
 endfunction
 function! deoplete#refresh() abort
   if exists('g:deoplete#_context')
-    let g:deoplete#_context.refresh = 1
     if get(g:deoplete#_context, 'event', '') ==# 'Manual'
       let g:deoplete#_context.event = 'Refresh'
     endif
--- a/autoload/deoplete/handler.vim
+++ b/autoload/deoplete/handler.vim
@@ -11,8 +11,8 @@ function! deoplete#handler#_init() abort
     autocmd CompleteDone * call s:complete_done()
     autocmd InsertCharPre * call s:on_insert_char_pre()
 
-    autocmd TextChangedI * call s:completion_check('TextChangedI')
-    autocmd InsertEnter * call s:completion_check('InsertEnter')
+    autocmd TextChangedI * call s:completion_begin('TextChangedI')
+    autocmd InsertEnter * call s:completion_begin('InsertEnter')
   augroup END
 
   for event in ['BufNewFile', 'BufRead', 'BufWritePost']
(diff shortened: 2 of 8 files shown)


dqgthb:
Correct me if I'm wrong. 
```
cd ~/.config/nvim/plugged/deoplete.nvim/
git checkout -b refresh
git pull refresh
```
and then I disabled the following line in my init.vim, just to test if this `refresh` commit can be a standalone solution to the problem.
`"let g:deoplete#auto_complete_delay = 250`

and, unfortunately, doing so could not solve the problem. As soon as I commented out the `auto_complete_delay = 250` line, the false behavior arose again. 

I can't tell precisely whether using both `refresh` branch and `auto_complete_delay = 250` would perfectly solve the problem. It is very difficult to measure the behavior as the frequency of the symptom already gets very low with `auto_complete_delay = 250` alone in my init.vim (like 1 time in 20 attempts or more). It is quite painful for both my brain and my fingers to trigger multiple snippets over and over dozens of times and to make sure it is not a human error. For now I will continue to put `auto_complete_delay = 250` in my init.vim and I'll definitely let you know if I get this behavior again.

By the way, this might be a bit off the topic but I found a number of other ways to prevent this problem in addition to `auto_complete_delay = 250`.
I hope this might give you a hint of what's going on. 
Each one was tested with other options (including `auto_complete_delay`) disabled.
## First 
Firstly, Adding this line also removed the symptom.
`let g:deoplete#auto_complete_start_length = 4`
I don't know the details of how Deoplete works but I suppose that the above line suppressed the activation of Deoplete (as most of my snippets are in 2-3 letters, such as fra<tab>,sqr<tab>,alp<tab>, etc.).

## Second (which is quite interesting)
Secondly, adding these lines toggles the symptom (!!).
```
let g:deoplete#ignore_sources = {}
let g:deoplete#ignore_sources._ = ['buffer']
let g:deoplete#ignore_sources = ['buffer']
" Basically what I was trying to do here is to disable buffer source. 
" Since the instructions for doing this were all different, 
" I just added all of them in my init.vim. 
" I hope you are not confused and please correct me if I did something wrong.
```
By 'toggle' I mean Vim alters its behavior of dealing with Ultisnips tabstops one after one.
As I said in my first issue report #444, the first usage of tabstop works normal, and the problem arises from the second usage and it continues the odd behavior. 

On the contrary, after adding these lines, it is like:
The first one behaves normal
The second behaves strange
The third behaves normal
The forth behaves strange
... 

So yeah, doesn't that sound fun? It might be related to how Deoplete gets its 'around' buffer.

Next time I would try to see what happens if I ignore both 'buffer' and 'around'.
```
let g:deoplete_ignore_sources = ['buffer', 'around']
```

Thanks for your prompt update, by the way. :)

Shougo:
Merged.

PR #467, opened 2017-04-14
[RFC] Fix async refresh on empty results

tweekmonster:
This fixes the issue mentioned in https://github.com/zchee/deoplete-jedi/pull/115

### DIFF
--- a/autoload/deoplete/handler.vim
+++ b/autoload/deoplete/handler.vim
@@ -3,6 +3,7 @@
 " AUTHOR: Shougo Matsushita <Shougo.Matsu at gmail.com>
 " License: MIT license
 "=============================================================================
+let s:changedtick = -1
 
 function! deoplete#handler#_init() abort
   augroup deoplete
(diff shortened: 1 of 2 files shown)


tweekmonster:
Actually, this only fixes the cases where `context['is_async'] = True` and candidates is empty.  It's still a good PR to deal with the deoplete-jedi's actual refresh issue.

There is still an issue with `b:changedtick` changing as mentioned in deoplete-jedi PR.  Video for reference: https://share.esdf.io/7dQUIEqnZt/

tweekmonster:
@Shougo Tracking `b:changedtick` only from `TextChangedI` seems to make this work correctly: https://share.esdf.io/nKsjNwv1Hd/

Shougo:
> Actually, this only fixes the cases where context['is_async'] = True and candidates is empty. It's still a good PR to deal with the deoplete-jedi's actual refresh issue.

The fix is incomplete.

tweekmonster:
> The fix is incomplete.

It was fixed by using tracking `b:changedtick` only from `TextChangedI`.  But, your other solution to simply ignore `b:changedtick` seems to work, too.

tweekmonster:
To clarify: The only missing part from 0a013fb is addressing `is_async` when `merged_results` is empty.

PR #468, opened 2017-04-14
Don't use previous results if empty

tweekmonster:
`use_previous_result()` should only return `True` if there are actually previous candidates to use.

### DIFF
--- a/rplugin/python3/deoplete/deoplete.py
+++ b/rplugin/python3/deoplete/deoplete.py
@@ -399,11 +399,12 @@ class Deoplete(logger.LoggingMixin):
                                                  attr, source_attr))
 
     def use_previous_result(self, context, result):
-        return (context['position'][1] == result['prev_linenr'] and
-                re.sub(r'\w*$', '', context['input']) ==
-                re.sub(r'\w*$', '', result['prev_input']) and
-                (not result['source'].is_volatile or
-                 context['input'].find(result['prev_input']) == 0))
+        return (result.get('candidates') and
+                (context['position'][1] == result['prev_linenr'] and
+                 re.sub(r'\w*$', '', context['input']) ==
+                 re.sub(r'\w*$', '', result['prev_input']) and
+                 (not result['source'].is_volatile or
+                  context['input'].find(result['prev_input']) == 0)))
 
     def is_skip(self, context, disabled_syntaxes,
                 min_pattern_length, max_pattern_length, input_pattern):


Shougo:
Why this feature is needed?
I don't understand.

tweekmonster:
This was causing an issue I noticed with deoplete-jedi where it returned empty results either because completions were taking too long or the server crashed.  There's no reasonable way to invalidate deoplete's cache.

Simple example:

```python
import random

class Source(Base):
    count = 0

    def gather_candidates(self, context):
        self.debug('gather')
        self.count += 1
        return [random.choice(['keyword', 'key_word'])] if random.randint(0, 10) else []
```

This gives a chance to get `keyword` or `key_word` or a rare random empty completion.  In order to retry a completion, I have to edit a _different line and undo it_ before I go back to the original line to try again.  But, there's still a chance that the problem will occur again.  In all cases, there's no way to reset the cache on the current line.

If a source returns zero candidates, it shouldn't be considered a valid cache entry.  At the very least, it should have a maximum age of a few seconds.  But, now that I think about it, the cache should be only valid while in insert mode.  Leaving insert mode should also reset the cache.

This cache implementation is **good** for throttling source execution when typing very fast or holding `<bs>`, but it shouldn't be a strict cache authority since deoplete is an agnostic framework that can't predict the needs of the sources.

Shougo:
Why don't you use `is_async`?

tweekmonster:
> Why don't you use is_async?

This was before I started working on the async refresh for deoplete-jedi.  But it still remains a problem for non-async sources because it's based on an assumption that the sources always have perfect operation.  For example, deoplete-clang2 will occasionally return nothing if `clang` is unable to parse the source because it crashes and produces a core dump.  Async refresh would be pointless in that case.

Since deoplete is filetype-neutral, I don't think it should maintain a non-expiring cache.  I think a short-term throttle cache is fine, though.  Caching should be under the source's control since they would actually understand the implications of using a cache for the current filetype and source location.

Shougo:
OK.  I get it.
But `Don't use previous results if empty` is not the solution.
`is_error` is needed for it.

Shougo:
It should be `None` instead of `is_error`.

tombeynon:
This PR seems to fix an issue I have with UltiSnips/Honza Snippets and Deoplete. It seems Deoplete is caching empty results

Type `rc`
// completion shown for `rc`, `rce`
// dd - clear line
Type `lt`
// no results

Move to another line (doesn't work on the same line, although this might be cache length)
Type `lt`
// completion shown for `lt`
// dd - clear line
Type `rc`
// no results

With this commit, all suggestions are shown as expected. I believe Deoplete caches the empty result from UltiSnips which isn't invalidated as intended, but I'm not sure.

Shougo:
> This PR seems to fix an issue I have with UltiSnips/Honza Snippets and Deoplete. It seems Deoplete is caching empty results

This is UltiSnips source issue.
The source must set `is_volatile=True` in the constructor.

Shougo:
I think UltiSnips source should not filter the candidates in `gather_candidates()`.
The filter is deoplete work.

PR #487, opened 2017-05-23
list of sources was burried in wiki

kimat:


README.md
```diff
@@ -44,6 +44,13 @@ You can enable Python3 interface with pip:
 
     pip3 install neovim
 
+## Sources
+
+deoplete will display completions via omnifunc by default.
+
+Here are some [completion sources](../../wiki/Completion-Sources) specifically made for deoplete.nvim.
```
Shougo: Please use https://github.com/Shougo/deoplete.nvim/wiki/Completion-Sources as the URI.
It should be absolute path.

PR #491, opened 2017-06-02
doc: Suppress pclose output to fix command-line window

Holzhaus:
`pclose` seems to cause an error when editing a command line window (the
window you get when pressing CTRL-F while editing the "`:`" command line):

> ```
> E11: Invalid in command-line window; <CR> executes, CTRL-C quits
> ```

This works around that problem by suppressing all error messages caused
by `pclose`.


Shougo:
Thanks

PR #509, opened 2017-07-12
Clarify documentation, add examples for 'dictionary' source

balta2ar:
Add link to 'ujihisa/neco-look' external source.

### DIFF
--- a/doc/deoplete.txt
+++ b/doc/deoplete.txt
@@ -571,7 +571,9 @@ tag							*deoplete-source-tag*
 		Note: It only supports UTF-8 encoding tag file.
 
 file							*deoplete-source-file*
-		This source collects keywords from local files.
+		This source collects keywords from local files. Specifically,
+		it completes file and directory names and paths, e.g.
+		/usr/lib/share.
 
 omni							*deoplete-source-omni*
 		This source collects keywords from 'omnifunc'.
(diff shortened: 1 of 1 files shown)


balta2ar:
@Shougo ping. Just bringing your attention.

Shougo:
Yes, I know it.
Please wait the merge.

doc/deoplete.txt
```diff
@@ -581,6 +583,18 @@ dictionary					*deoplete-source-dictionary*
 		This source collects |g:deoplete#keyword_patterns| keywords
 		from 'dictionary'.
```
Shougo: The 'dictionary' option must be buffer local.
I think it should be documented.

doc/deoplete.txt
```diff
@@ -581,6 +583,18 @@ dictionary					*deoplete-source-dictionary*
...
 
+		" Examples:
+		" Sample configuration for dictionary source with multiple
+		" dictionary files.
+		setlocal dictionary+=/usr/share/dict/words
+		setlocal dictionary+=/usr/share/dict/american-english
```
Shougo: So, the `setlocal dictionary` is true.  And `set dictionary` is wrong.

balta2ar: Hm... I think it worked even with `set dictionary`. What's wrong with using non-buffer-local set up?

Shougo: dictionary source only checks buffer local dictionary.
But set change the global and buffer local value.
It may be works.  But it is not intended.

balta2ar: Yes, but why only check buffer local dictionary? So that you can use different dictionaries for different buffers with different file types?

Shougo: > So that you can use different dictionaries for different buffers with different file types?

Yes.

PR #522, opened 2017-08-06
Update docs after changing the default matchers

svanharmelen:
I guess this is the root cause of my misunderstanding of the new experience as I just noticed you also changed (removed the `matcher_length`) the default in commit 330f356

So this doc update should represent that change and make it a bit more clear :wink:


Shougo:
Please resolve the conflict.

Shougo:
Merged.

svanharmelen:
Thx!

PR #538, opened 2017-09-04
improve doc

e-kwsm:
* add the `rank`s of `deoplete-sources`
* add explanation of `mark`, `description`, and `is_initialized` to `deoplete-source-attributes`

### DIFF
--- a/doc/deoplete.txt
+++ b/doc/deoplete.txt
@@ -572,26 +572,38 @@ buffer						*deoplete-source-buffer*
 		Note: It takes time to get the candidates in the first time if
 		you want to edit the large files(like Vim 22000 lines eval.c).
 
+		rank: 100
+
 member						*deoplete-source-member*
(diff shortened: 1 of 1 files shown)


doc/deoplete.txt
```diff
@@ -793,6 +815,11 @@ is_bytepos
 		Because Vim script string uses byte position.  Python string
 		uses character position.
 
+					*deoplete-source-attribute-is_initialized*
+is_initialized
+		(Bool)				(Optional)
+		If it is True, |deoplete-source-attribute-on_init| is not called.
```
Shougo: It is needed?
It is just used internally.

e-kwsm: None of the sources listed in [wiki](https://github.com/Shougo/deoplete.nvim/wiki/Completion-Sources) seem to use it; removed.

PR #540, opened 2017-09-13
Fix: manual_complete trapped in state check

halftan:
`deoplete#initialize()` will always return 1 after deoplete is enabled, so `manual_complete` will always fall into the if clause.


Shougo:
Thanks.  But I think my fix is better.

PR #548, opened 2017-09-29
Update README.md

jrowlingson:
I may be interpreting the doc incorrectly, but, it looks like the completion method defaults to [`'complete'`](https://github.com/Shougo/deoplete.nvim/blob/1c33e9a3b118208d4c48bf3ef0a1ad77e16bc0a1/autoload/deoplete/init.vim#L111)

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,7 +48,7 @@ You can enable Python3 interface with pip:
 
 ## Sources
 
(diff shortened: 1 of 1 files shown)


Shougo:
Thanks

PR #555, opened 2017-10-16
Probably a much faster fuzzy matcher

BurningLutz:
Instead of regexp, I use two cursors to determine whether string B is a non-continuous sub string of String A.

### DIFF
--- a/rplugin/python3/deoplete/filter/matcher_fuzzy.py
+++ b/rplugin/python3/deoplete/filter/matcher_fuzzy.py
@@ -4,10 +4,8 @@
 # License: MIT license
 # ============================================================================
 
-import re
 from .base import Base
-from deoplete.util import (
-    fuzzy_escape, binary_search_begin, binary_search_end)
+from deoplete.util import fuzzy_match
 
 
 class Filter(Base):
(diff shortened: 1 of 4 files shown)


Shougo:
It is not so bad...
Please profile the improvement if it is much faster.
And why you have removed the binary search feature?

rplugin/python3/deoplete/filter/matcher_fuzzy.py
```diff
@@ -18,24 +16,18 @@ def __init__(self, vim):
...
-        if context['is_sorted']:
-            begin = binary_search_begin(
-                context['candidates'], complete_str[0])
-            end = binary_search_end(
-                context['candidates'], complete_str[0])
-            if begin < 0 or end < 0:
-                return []
-            candidates = context['candidates'][begin:end+1]
-        else:
-            candidates = context['candidates']
```
Shougo: Why you have removed the binary search feature?
Please describe it.

BurningLutz: Hmmm...I'm not sure what is the binary search for. I thought it's for performance purpose, isn't it?

And I thought there is a bug of the binary search in the current fuzzy matcher. It doesn't reflect to the `ignorecase` option correctly.
Thus, the test case below will fail:
```python
    ctx = {
        'complete_str': 'fobr',
        'ignorecase': True,
        'camelcase': False,
        'is_sorted': False,
        'candidates': [
            { 'word': 'FooBar' },
            { 'word': 'foobar' },
            { 'word': 'aFooBar' },
            { 'word': 'fooBar' },
        ]
    }
    assert f.filter(ctx) == [
        { 'word': 'FooBar' },
        { 'word': 'foobar' },
        { 'word': 'fooBar' },
    ]
```
I will provide profiling results later, hmm.... maybe tomorrow.

Shougo:
> Hmmm...I'm not sure what is the binary search for. I thought it's for performance purpose, isn't it?

Yes.

Shougo:
> Thus, the test case below will fail:

In ignorecase, the order is undefined.

Shougo:
> I will provide profiling results later, hmm.... maybe tomorrow

Please compare the huge candidate results.

BurningLutz:
I'm afraid I can not provide profiling results because the f*cking GFW blocked gist.github.com and banned my VPN.

Shougo:
> I'm afraid I can not provide profiling results today because the f*cking GFW blocked gist.github.com and banned my VPN.

Oh,...

BurningLutz:
@Shougo 
I think I've made a mistake when doing profiling before I wrote the new fuzzy matcher.I tried to call `filter` many times to test its performance but it's not the case.

The real situation is that, `filter` is not called frequently so the most performance consuming `re.compile` affects performance little. And when doing candidates filtering, my `fuzzy_match` is slower than `p.match` at the most of time.

So I decide to close this PR.

Shougo:
OK.

But can you update the tests?
Your tests method seems better.

BurningLutz:
@Shougo 
OK. I'll do that.

PR #559, opened 2017-10-23
add tests for matcher_fuzzy and matcher_full_fuzzy

BurningLutz:
@Shougo I extract `test_matcher_fuzzy` and `test_matcher_full_fuzzy` into individual files to make it more declarative.

Do you agree with it?


Shougo:
Merged.

Issue #577, opened 2017-11-11
Deoplete appears to fail to initialize in vim 8.


letharion:
# Problems summary
Deoplete prints "[deoplete] function deoplete#enable[1]..deoplete#initialize[1]..deoplete#init#_initialize[9].....rp#core#jobstart[2]..yarp#pyx#init[13]..yarp#core#serveraddr[1]..neovim_rpc#serveraddr" to the bottom of the screen, doesn't autocomplete.

## Expected
No error message, autocompletion.

## Environment Information

 * deoplete version(SHA1):
3c38321017ae888f7807ff61ae98e0f21239bbd2

 * OS:
Ubuntu Linux

 * neovim/Vim version:
VIM - Vi IMproved 8.0 (2016 Sep 12, compiled Nov  6 2017 11:22:14)
Included patches: 1-1272
Modified by pkg-vim-maintainers@lists.alioth.debian.org
Compiled by pkg-vim-maintainers@lists.alioth.debian.org

## Provide a minimal init.vim/vimrc with less than 50 lines (Required!)
```vim
if &compatible
  set nocompatible               " Be iMproved
endif

" Required:
set runtimepath+=/home/ubuntu/.vim/plugins/repos/github.com/Shougo/dein.vim

" Required:
if dein#load_state('/home/ubuntu/.vim/plugins')
  call dein#begin('/home/ubuntu/.vim/plugins')

  " Let dein manage dein
  " Required:
  call dein#add('/home/ubuntu/.vim/plugins/repos/github.com/Shougo/dein.vim')

  " Add or remove your plugins here:
  call dein#add('Shougo/neosnippet.vim')
  call dein#add('Shougo/neosnippet-snippets')

  " You can specify revision/branch/tag.
  call dein#add('Shougo/vimshell', { 'rev': '3787e5' })

  call dein#add('Shougo/deoplete.nvim')
  if !has('nvim')
    call dein#add('roxma/nvim-yarp')
    call dein#add('roxma/vim-hug-neovim-rpc')
  endif

  " Required:
  call dein#end()
  call dein#save_state()
endif

" Required:
filetype plugin indent on
syntax enable

" If you want to install not installed plugins on startup.
if dein#check_install()
  call dein#install()
endif
let g:deoplete#enable_at_startup = 1
```

## The reproduce ways from neovim/Vim starting (Required!)
Starting vim is enough to generate the error message

However, full details of how I tested this is:
mkdir /tmp/deopletetest
[...]

PR #592, opened 2017-12-05
Update README.md

letharion:
I propose a minor change to the wording of the python 3 requirement. I interpreted "you're done", as being done with the whole requirements section, and thus I stopped reading the section at this point. This meant I never installed nvim-yarp or vim-hug-neovim-rpc, which in turn lead me to open an issue (https://github.com/Shougo/deoplete.nvim/issues/577).

Perhaps a different wording will make this less likely to become an issue in the future.

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,7 +50,7 @@ endif
 ## Requirements
 
 deoplete requires Neovim or Vim8 with if\_python3.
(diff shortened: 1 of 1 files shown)


Shougo:
Merged.

PR #607, opened 2018-01-12
Parse functions in Tag source

tweekmonster:
Makes the tag source a little more useful.

screen-shot-2018-01-12-17-19-40

### DIFF
--- a/rplugin/python3/deoplete/source/tag.py
+++ b/rplugin/python3/deoplete/source/tag.py
@@ -5,6 +5,8 @@
 # License: MIT license
 # ============================================================================
 
+import re
+
 from .base import Base
 
 from collections import namedtuple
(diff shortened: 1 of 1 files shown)


Shougo:
Merged.

tweekmonster:
Thanks!

kristijanhusak:
@tweekmonster @Shougo i have a small issue with this change, nothing really big or important, but i would like to know if there's a way to solve it or i can create a PR with the fix.
I'm using Universal Ctags for Node JS, and it generates these types of tags:

```
CustomerValidator	lib/domain/user/signup_validator.js	/^const CustomerValidator = require('..\/..\/customer\/customer_validator');$/;"	C
```

When i try to autocomplete `CustomerValidator`, i get this:
screenshot

Which turns out to be a bit confusing. I would like to make an exception for this case so i the default autocomplete:
screenshot2

Is there any way to do this without actually changing this python code? I assume not.
@Shougo if it's ok with you to add this fix, i'll create a PR.

Shougo:
OK.  It should be fixed.
The PR is wellcome.

PR #612, opened 2018-01-24
Performance improvement: when using bufnr use % not .

dncnmcdougall:
I'm inclined to use A LOT of buffers. Sometimes when I make a change and write it takes a whole second. This is really irritating. 
I did a quick performance profile and it seems that the culprit is using bufnr('.').
I've changed this to bufnr('%') as that seems to be the expected behaviour.


Shougo:
Thanks

PR #625, opened 2018-02-04
wait to kill python processes.

RyoSusami:
Add wait to kill python processes.

### DIFF
--- a/autoload/deoplete/util.vim
+++ b/autoload/deoplete/util.vim
@@ -218,7 +218,7 @@ function! deoplete#util#rpcnotify(event, context) abort
     return ''
   endif
   call s:notify(a:event, a:context)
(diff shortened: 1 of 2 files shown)


Shougo:
Thanks.

blueyed:
Doesn't this only wait for the first child to exit?

blueyed:
And why is this necessary after all?

Shougo:
> Doesn't this only wait for the first child to exit?

I will fix it later.

> And why is this necessary after all?

Because neovim exits without waiting subprocess.
Note: Vim8 does not have the problem.

blueyed:
I think https://github.com/Shougo/deoplete.nvim/pull/631 might take care of that (childs exit when stdin gets closed).

Shougo:
OK.  But it is big change.
I must test it before merge.

Shougo:
More testers are needed.

PR #629, opened 2018-02-05
logger: add process/PID to log_format

blueyed:
This is useful to have given that there are child processes now by
default, and also when logging to the same file from multiple sources
(Neovim instances).


Shougo:
OK.  Merged.

PR #631, opened 2018-02-05
Refactor Child.main to handle EOF

blueyed:
I've seen child processes being left over after quitting Neovim, looping
in `read()`.
This appears to fix it.

Maybe this is what https://github.com/Shougo/deoplete.nvim/pull/625
tried to address also?!

### DIFF
--- a/rplugin/python3/deoplete/child.py
+++ b/rplugin/python3/deoplete/child.py
@@ -49,31 +49,36 @@ class Child(logger.LoggingMixin):
             unicode_errors='surrogateescape')
 
     def main(self):
-        for child_in in self._read():
-            self.debug('main_loop: begin')
-            name = child_in['name']
-            args = child_in['args']
-            queue_id = child_in['queue_id']
-            self.debug('main_loop: %s', name)
+        while True:
+            feed = sys.stdin.buffer.raw.read(102400)
+            if feed == b'':
+                # EOF
+                return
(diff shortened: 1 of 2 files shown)


rplugin/python3/deoplete/child.py
```diff
@@ -49,31 +49,36 @@ def __init__(self, vim):
...
-            elif name == 'on_event':
-                self._on_event(args[0])
-            elif name == 'merge_results':
-                self._write(self._merge_results(args[0], queue_id))
-
-    def _read(self):
-        self._unpacker.feed(sys.stdin.buffer.read(1))
-        return self._unpacker
+        while True:
+            feed = sys.stdin.buffer.raw.read(102400)
```
blueyed: This also reads up to 100kb (I am seeing ~10kb at minimum here).
Reading this byte per byte seems to be very wasteful.

Shougo:
It is very big change.
It needs to be testing.

I will test it later.

Shougo:
More testers are needed.

lambdalisue:
I haven't test yet but code looks quite simple and efficient. LGTM

Shougo:
I will test it later.
Please wait.

Shougo:
It seems works.

I will merge it later.

Shougo:
Merged.

PR #641, opened 2018-02-07
Process.enqueue_output: use raw read

blueyed:
This will close the Process/Thread automatically on EOF, resulting in
subprocesses to quit without waiting for them.

### DIFF
--- a/rplugin/python3/deoplete/parent.py
+++ b/rplugin/python3/deoplete/parent.py
@@ -62,7 +62,6 @@ class Parent(logger.LoggingMixin):
     def on_event(self, context):
         if context['event'] == 'VimLeavePre':
             self._stop_process()
-            self._vim.vars['deoplete#_stopped_processes'] += 1
         self._put('on_event', [context])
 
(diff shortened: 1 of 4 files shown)


rplugin/python3/deoplete/parent.py
```diff
@@ -62,7 +62,6 @@ def merge_results(self, context):
     def on_event(self, context):
         if context['event'] == 'VimLeavePre':
             self._stop_process()
-            self._vim.vars['deoplete#_stopped_processes'] += 1
         self._put('on_event', [context])
```
blueyed: Side note: `_stop_process` should get called at the end likely, since otherwise the `_put` will not do anything.
I assume it is good to propagate the event to childs, and afterwards stop?!

blueyed: @Shougo 
?

Shougo: OK.  It should be.
I will fix it.

PR #648, opened 2018-02-08
Skip creating tag abbreviation for definitions that contain js require().

kristijanhusak:
Fixes issue mentioned in comment https://github.com/Shougo/deoplete.nvim/pull/607#issuecomment-364050997

### DIFF
--- a/rplugin/python3/deoplete/source/tag.py
+++ b/rplugin/python3/deoplete/source/tag.py
@@ -22,6 +22,7 @@ class Source(Base):
 
         self.name = 'tag'
         self.mark = '[T]'
+        self.abbr_blacklist = ['require(']
 
         self._cache = {}
 
@@ -54,7 +55,11 @@ class Source(Base):
                     if not cols or cols[0].startswith('!_'):
                         continue
                     i = cols[2].find('(')
-                    if i != -1 and cols[2].find(')', i+1) != -1:
+                    blacklisted = any(
+                        item in cols[2] for item in self.abbr_blacklist
+                    )
+                    if (i != -1 and cols[2].find(')', i+1) != -1
+                            and not blacklisted):
                         m = re.search(r'(\w+\(.*\))', cols[2])
                         if m:
                             items.append({'word': cols[0],


tweekmonster:
@kristijanhusak I don't like the way a blacklist is being used here.  `require()` isn't exclusive to javascript.  I think the ctag attributes will need to be parsed to determine if it's actually a function: http://docs.ctags.io/en/latest/format.html

kristijanhusak:
@tweekmonster what other language uses `require()`?

I don't think we can get an attribute that will tell us that it's javascript definition, correct me if i'm wrong.

One thing we can do is determine the extension of the file in the tag definition, and act according to that.

tweekmonster:
> @tweekmonster what other language uses require()?

PHP.  I could also create a `require()` function in C or most any other language for that matter.  It also looks like the blacklist would match `not_require()` as well.

> I don't think we can get an attribute that will tell us that it's javascript definition, correct me if i'm wrong.

You can get the filetype from the `context` variable.  But, I was referring to ctag's attributes.

From your example:

```
CustomerValidator	lib/domain/user/signup_validator.js	/^const CustomerValidator = require('..\/..\/customer\/customer_validator');$/;"	C
```

The `C` at the end is denoting that the tag is a constant.  For functions, it would be `f`.  But, there's more to the ctag format that might not make this a straight-forward evaluation of the last character.

The other thing is that only the first 3 fields (before the `;"`) is actually required.  I wasn't sure if universal-ctags always added the extra fields, so I just stuck with regex matching, which I'll admit was pretty naive.

kristijanhusak:
@tweekmonster You are right, i didn't think about someone creating that function themselves.
I know for those attributes, but i don't think they are reliable enough.

What are your suggestions?

tweekmonster:
Created a PR that should address your issue and incidentally adds a little more useful info to other tag symbols.  If people have tags that are generated without the extra tagfields, then no special treatment.

kristijanhusak:
Closing in favor of https://github.com/Shougo/deoplete.nvim/pull/650

PR #650, opened 2018-02-08
Parse ctags tagfield to get symbol kind

tweekmonster:
ref #607 #648

ping @kristijanhusak


Shougo:
Please test @kristijanhusak.
I think it is better solution.

kristijanhusak:
@Shougo I agree.

@tweekmonster thanks, this is much better solution.

Shougo:
I will merge it later.

Shougo:
Merged.

PR #661, opened 2018-02-21
deoplete#util#print_error: handle optional name

blueyed:
This is used from the Python logger to pass through the logger's name,
which is useful when logging from sources (e.g. deoplete-jedi).


Shougo:
Thanks

PR #668, opened 2018-02-23
Remove load_external_module

blueyed:


Shougo:
No.  The function is used for several sources.
It is needed.

Shougo:
Example:

```python
load_external_module(__file__, 'sources/deoplete_go')
from cgo import cgo
from stdlib import stdlib
```

PR #685, opened 2018-03-14
buffer source: skip cache with BufNew

blueyed:
_make_cache acts on the current buffer, which usually is not the new
buffer with BufNew, and the buffer is supposed to be empty anyway.


Shougo:
No need.

PR #700, opened 2018-03-23
reverse order of expands in deoplete.util.expand()

deathlyfrantic:
With the previous order of expand calls, it's possible to have an unexpanded `~`:

```
>>> os.environ.get('FOO')
'~/src/vim'
>>> path.expandvars(path.expanduser('$FOO'))
'~/src/vim'
>>> path.expanduser(path.expandvars('$FOO'))
'/Users/zandr/src/vim'
```


Shougo:
Merged.

PR #711, opened 2018-03-31
Source.events: default to None and improve docs

blueyed:
`events = []` should allow for meaning "never call on_event".


Shougo:
Merged.

PR #712, opened 2018-03-31
(Re)move/inline SubprocessProtocol callbacks

blueyed:
This makes the code clearer / centralizes it.


blueyed:
Has this been done like this because of https://github.com/neovim/python-client/blob/debcde029eceef10c5ece1bc430452d17180e62b/neovim/api/nvim.py#L40-L44 ?

PR #716, opened 2018-04-01
Base: move most of attribute initialization out of __init__

blueyed:


Shougo:
Merged.

lambdalisue:
I don't get why this change is required. All instance including sub-class would share the insatnce of `matchers`, `sorters`, `converters`, `filetypes`, and `disabled_syntaxes` so I think it's a kind trap.

For example, are you sure that the following behavior is what you expect?

```python
class A:
    foo = []

class B(A):
    def __init__(self):
        self.foo.append('Hello')

class C(A):
    def __init__(self):
        self.foo.append('Goodbye')

print(C().foo)
print(B().foo)
print(A().foo)

# ['Goodbye']
# ['Goodbye', 'Hello']
# ['Goodbye', 'Hello']
```

Shougo:
I have reverted it.

Reason: @lambdalisue 's comment.

blueyed:
Yes, sorry - it does not work for mutable types.
Thanks for reverting already.

PR #719, opened 2018-04-02
s:completion_begin: improve check for omni completion

blueyed:
Move conditional checks out of the loop(s).


Shougo:
Please fix the conflict.

Shougo:
I have pushed the better version.

blueyed:
Thanks!

PR #721, opened 2018-04-02
Fixed tags binary search

vitaly:
The previous algorithm was quite broken.
It would skip the lines, e.g. last line was (always?) skipped, also
it would usually skip first couple of lines matching pattern.

Also the new algorithm does 2 times less line reads, as it reads only
one line per seeking iteration.


Shougo:
Merged.

PR #722, opened 2018-04-02
Revert "Base: move most of attribute initialization out of __init__"

Shougo:
Reverts Shougo/deoplete.nvim#716


Shougo:
Reverted.

PR #742, opened 2018-04-25
advertise usage of 'deoplete#custom#option()'

kevinushey:
This change makes it (IMHO) a bit more clear that `deoplete#custom#option()` is the recommended API for setting options.


Shougo:
Merged

PR #750, opened 2018-05-03
Added another command for closing preview window

enkeyz:
With this command, preview window only close, if you exit INSERT mode.


Shougo:
My version is better.

PR #768, opened 2018-05-25
Add instructions for tab completion to README

jacklenox:
I had to dig around to find the best way to do this. I suspect this might be commonly desired from users and therefore would be useful in the README.


Shougo:
I don't think the information is useful in README.
`<Tab>` mapping is already included in example section in doc.

hasufell:
I found this useful

PR #787, opened 2018-07-05
Handle SourceInitError when gathering results

blueyed:
Handles `SourceInitError` in `_gather_results` and
`_gather_async_results` in the same way as when loading a source (msg
and no traceback).

This is meant to be used by deoplete-jedi, which might only know about
failing workers when results are gathered for the first time.


Shougo:
Merged.

PR #792, opened 2018-07-10
context: drop "runtimepath"

blueyed:
It is a lot of overhead (especially with many plugins installed).

`vim.options['runtimepath']` can be used instead if necessary.

Adjusts cpsm matcher and improves its setup.


blueyed:
If backward compatibility is important than we could have a custom dict that looks up `'runtimepath'` dynamically on request, and logs some deprecation warning.

blueyed:
For B/C also `set_source_attributes` could be used to set it as an attribute.

Shougo:
OK

blueyed:
Great.. reduces the size of the context by 70% or so for me.. especially important when printing/logging it.

PR #793, opened 2018-07-10
_check_recache: set _custom to not call set_source_attributes always

blueyed:
Not sure if/why this is needed after all, but at least it should update `self._custom`, which appears to be only used for this check.


Shougo:
Merged.

blueyed:
Thanks!

PR #796, opened 2018-07-10
Deoplete.__init__: fall back to single process

blueyed:
This situation can be handled easily, no need to stop working because of
it.


Shougo:
OK

PR #798, opened 2018-07-10
Split Parent into SingleParent/MultiParent

blueyed:
This cleans up responsibilities between code, and makes it easier to
grasp what gets done in what case.

This also avoids passing `context` to `Parent` to get the number of
processes from it.

I've also removed the `cwd=context['cwd']` with
`self._vim.loop.subprocess_exec` for MultiParent - assuming that it will
default to the current one.
If that assumption is wrong it should be fixed for
`SingleParent`/`Child` then also/instead.

### DIFF
--- a/rplugin/python3/deoplete/deoplete.py
+++ b/rplugin/python3/deoplete/deoplete.py
@@ -4,13 +4,13 @@
 # License: MIT license
 # ============================================================================
 
-from deoplete import logger
-from deoplete.parent import Parent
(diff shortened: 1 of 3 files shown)


Shougo:
Please fix the conflict.

blueyed:
Fixed.

Shougo:
OK.  I will merge it later.

Shougo:
Merged.

PR #799, opened 2018-07-10
Clean up logging a bit

blueyed:
The current debug logging is very spammy, and this makes it more useful.

### DIFF
--- a/rplugin/python3/deoplete/child.py
+++ b/rplugin/python3/deoplete/child.py
@@ -56,19 +56,15 @@ class Child(logger.LoggingMixin):
                 return
 
             self._unpacker.feed(feed)
-            self.debug('_read: %d bytes', len(feed))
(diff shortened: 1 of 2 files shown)


blueyed:
Thanks for merging.

However, commits with a `fixup!` prefix are meant to be fixed-up, e.g. by using `git rebase -i --autosquash`, or just by using squash-merging from the Github UI.  Just for information.. - I hope we can work this out in the future.. :)

blueyed:
I see.
However it can be squashed-merged from Github, and therefore I tend to add TODOs instead, but you should just check before merging.
In this case there was even a CI failure - which indicates WIP somehow.

(I like to avoid rebasing/force-pushing if not necessary, i.e. if a fixup fixes the build, then it can be squash/rebase-merged from there)

PR #802, opened 2018-07-11
Measure code coverage and upload to codecov

blueyed:


blueyed:
I plan to look into getting coverage for Vimscript also, but that involves work on covimerage to wrap themis.

blueyed:
@Shougo 
You can browse the coverage at https://codecov.io/gh/Shougo/deoplete.nvim/src/5abc7d295edc62cc2c86c8b7792f22e6d3dd29a3/rplugin/python3/deoplete

We can configure or disable the codecov comment, but I think it is good in general, but the following config would make it less noisy:

```
coverage:
  status:
    project: false
    patch: true
    changes: true

comment:
  layout: "diff"
```

This should be configured in the Codecov settings / UI then (I would rather not add a file to repo for it, but it is possible, too).

Shougo:
OK.

Issue #769, opened 2018-05-26
cpoptions breaks deoplete


petong:
**Warning:  I will close the issue without the minimal init.vim and the reproduction instructions.**

# Problems summary
setting cpoptions in init.vim breaks deoplete. typing two consecutive characters in input mode results in ```<C-r>=deoplete#mapping#_complete()<CR>```

## Expected
deoplete should complete after typing two consecutive characters

## Environment Information

 * deoplete version(SHA1): 165e693cbb1198569954b2a01f7ba56026bf7ddf

 * OS: OSX 10.13.5

 * neovim/Vim `:version` output: 
```
NVIM v0.2.2
Build type: Release
LuaJIT 2.0.5
Compilation: /usr/local/Homebrew/Library/Homebrew/shims/super/clang -Wconversion -U_FORTIFY_SOURCE -D_FORTIFY_SOURCE=1 -DNVIM_MSGPACK_HAS_FLOAT32 -DNVIM_UNIBI_HAS_VAR_FROM -DNDEBUG -DMIN_LOG_LEVEL=3 -Wall -Wextra -pedantic -Wno-unused-parameter -Wstrict-prototypes -std=gnu99 -Wimplicit-fallthrough -Wvla -fstack-protector-strong -fdiagnostics-color=auto -DINCLUDE_GENERATED_DECLARATIONS -I/tmp/neovim-20180209-92407-udkzoo/neovim-0.2.2/build/config -I/tmp/neovim-20180209-92407-udkzoo/neovim-0.2.2/src -I/usr/local/include -I/usr/local/include -I/usr/local/include -I/usr/local/include -I/usr/local/include -I/usr/local/include -I/usr/local/opt/gettext/include -I/usr/include -I/tmp/neovim-20180209-92407-udkzoo/neovim-0.2.2/build/src/nvim/auto -I/tmp/neovim-20180209-92407-udkzoo/neovim-0.2.2/build/include
Compiled by brew@HighSierra.local

Features: +acl +iconv +jemalloc +tui
See ":help feature-compile"

   system vimrc file: "$VIM/sysinit.vim"
  fall-back for $VIM: "/usr/local/Cellar/neovim/0.2.2_1/share/nvim"```

 * `:checkhealth` or `:CheckHealth` result(neovim only):

## Provide a minimal init.vim/vimrc with less than 50 lines (Required!)

```
call plug#begin('~/.config-test/nvim/plugged')
[...]

PR #806, opened 2018-07-12
Do not trigger <c-e> for already visible pum

blueyed:
It was added in d8a8670 [1], without any comment/reference.

This fixes an issue, where deoplete reverts an already selected item from
the popup menu (using normal `<C-p>`), when new completions arrive.

In my case the later completions are coming from deoplete-jedi (slower),
but it can also be reproduced with the around source:

Given:

    $shims_base_dir
    $shims_

Place the cursor in the 2nd line, type `A` and `Ctrl-P` fast: The Ctrl-P
will select `$shims_base_dir` then.
When deoplete sets the completions `$shims_` gets added additionally (which is a bug
in the around source it seems).

With this change `$shims_` will show up shortly in the completion list,
but `$shims_base_dir` will still be inserted/selected.

This is obviously caused by the `<C-e>` that deoplete sends itself, and
should not be necessary at all.

This also removes the `deoplete#mapping#_dummy_complete` wrapper.

I could not reproduce https://github.com/Shougo/deoplete.nvim/issues/769
with `set cpoptions=c`.  Maybe it was related to the conditional that
was used before (see 86eeb17).

1: https://github.com/Shougo/deoplete.nvim/commit/d8a8670

### DIFF
--- a/autoload/deoplete/mapping.vim
+++ b/autoload/deoplete/mapping.vim
@@ -5,8 +5,7 @@
 "=============================================================================
 
 function! deoplete#mapping#_init() abort
-  " Note: The dummy function is needed for cpoptions bug in neovim
-  inoremap <expr><silent> <Plug>_ deoplete#mapping#_dummy_complete()
+  inoremap <expr><silent> <Plug>_ "\<C-r>=deoplete#mapping#_complete()\<CR>"
 endfunction
 
 function! deoplete#mapping#_completefunc(findstart, base) abort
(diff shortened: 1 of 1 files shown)


Shougo:
Hm.  OK.  I will merge it.

teto:
with this patch, I get https://github.com/Shougo/deoplete.nvim/issues/769 again; had to revert to f012054fb6a07801c34614a036e4fd9171b79190 . `set cpoptions?` is empty

blueyed:
I see.  So the dummy function would be needed in this case again.
(but without the `<c-e>` part then)
Or is there a better option?

Why do you use this `cpoptions` setting?
Doesn't it break a lot of other plugins, too?

Shougo:
Fixed.

PR #809, opened 2018-07-19
Fix rplugin paths so they are OS agnostic (make rplugins work on Windows)

dhazel:
**The problem**: When attempting to load deoplete rplugins on Windows7 the path separators are currently a mix of forward slashes and back slashes on windows, which makes the paths unreachable.

Example:

    C:\path\to\vim\vimfiles\plugged\deoplete-omnisharp\rplugin/python3/deoplete\sources\cs.py

**The fix**: This changeset revises the code to use the path separators for the entire path that the current os supports.


Shougo:
OK.

PR #817, opened 2018-07-31
Correct event handling of 'buffer' source

fgrsnau:
The 'buffer' source currently has some non-obvious behaviour. It refreshes the cache for a given buffer when writing it to a file and on `InsertEnter`. The last thing happens only when there is no cache for buffer before. As the buffer caches get never invalidated, essentially the `InsertEnter` event hook only works at most once during a session.

All the above means, that the cache for a buffer is never populated if the buffer gets never modified. If you have lots of buffers open and want to edit only a single file, you don't get any meaningful completions.

I changed the behaviour as follows:

1.  Add a new autocmd hook for `BufReadPost`.
2. Refresh the cache for a buffer unconditionally on `BufReadPost` and `BufWritePost` and don't hook other events.

This results in having the buffer source providing completions for the on-disk state of every open buffer which is what the average user would expect :)

### DIFF
--- a/autoload/deoplete/handler.vim
+++ b/autoload/deoplete/handler.vim
@@ -12,7 +12,7 @@ function! deoplete#handler#_init() abort
     autocmd InsertLeave * call s:completion_timer_stop()
   augroup END
 
-  for event in ['InsertEnter', 'BufWritePost']
+  for event in ['InsertEnter', 'BufReadPost', 'BufWritePost']
     call s:define_on_event(event)
   endfor
 
--- a/rplugin/python3/deoplete/source/buffer.py
+++ b/rplugin/python3/deoplete/source/buffer.py
@@ -16,7 +16,7 @@ class Source(Base):
 
         self.name = 'buffer'
         self.mark = '[B]'
-        self.events = ['InsertEnter', 'BufWritePost']
+        self.events = ['BufReadPost', 'BufWritePost']
         self.vars = {
             'require_same_filetype': True,
         }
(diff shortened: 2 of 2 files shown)


Shougo:
OK.  Merged.

Shougo:
Hm....

It has https://github.com/Shougo/denite.nvim/issues/457 problem.

Can you fix it?

If not, I may revert the PR.

fgrsnau:
I didn't know about denite but from what I've seen so far it looks quite useful. :)

Unfortunately, I don't know how to reproduce the issue. I don't know what "switch to normal mode" and "press p for preview" means. In normal mode "p" means paste. The documentation says that `<C-V>` is mapped to preview and I can toggle the preview a dozen of times without any problems.

I could imagine that the code in the buffer source is buggy, because it's buffer cache is never freed. Let's say we open 1000 buffers and close 999 of them, we still keep 1000 entries in the `_buffers` cache. I think, hooking `BufDelete` or `BufWipeout` is a good idea. But as far as I see denite is not creating a new buffer for each preview, so it's probably unrelated.

Shougo:
Don't mind.  I have fixed the problem.

> I could imagine that the code in the buffer source is buggy, because it's buffer cache is never freed. Let's say we open 1000 buffers and close 999 of them, we still keep 1000 entries in the _buffers cache.

I have fixed the problem.

Shougo:
> I could imagine that the code in the buffer source is buggy, because it's buffer cache is never freed. Let's say we open 1000 buffers and close 999 of them, we still keep 1000 entries in the _buffers cache. I think, hooking BufDelete or BufWipeout is a good idea. But as far as I see denite is not creating a new buffer for each preview, so it's probably unrelated.

Unfortunately, this is not good idea.

Because: https://github.com/Shougo/deoplete.nvim/issues/818

fgrsnau:
I would argue that it is still a good idea, but that your latest commit did not implement it quite correctly ;) I also noticed the crashing due to a `IndexError`/`KeyError`s. If I find the time send another PR with a better tested alternative.

PR #852, opened 2018-09-28
Change README options reference

Istanful:
# Change README options reference
To read about available options the README referenced `deoplete-options`
which is not available. It is now changed to `deoplete-variables`. :)


Shougo:
`:help deoplete-options` is available.

Shougo:
Don't mind.

PR #864, opened 2018-10-24
Ignore header of changes command

tex:
$SUBJ$

I don't like the 'col', 'line', 'text' coming as header of changes command...


Shougo:
OK

PR #870, opened 2018-11-07
Improve document consistency and clarity

brianvanburken:
I was reading the documentation with `:help deoplete-optiosn`, as a new user, and found some inconsistency issues and unknown units.

### DIFF
--- a/doc/deoplete.txt
+++ b/doc/deoplete.txt
@@ -104,27 +104,27 @@ The set of available options follows.
 async_timeout
 		The timeout of asynchronous completion.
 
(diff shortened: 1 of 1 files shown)


kiryph:
In my opinion the change `default value: v:false` to `default value: false` is misleading. There is no  keyword `false` (or `true`) in vim. A beginner could think to use `false` or `true`. This does not work. Either you use `0/1` or the special value `v:false/v:true`.

Shougo:
Hm.  OK.  I will replace it later.

PR #876, opened 2018-11-18
Import pynvim if find pynvim module

h-michael:
```
 ~> nvim --version
NVIM v0.3.2-893-g180b50ddd
Build type: RelWithDebInfo
LuaJIT 2.0.5
Compilation: /usr/bin/cc -march=x86-64 -mtune=generic -O2 -pipe -fstack-protector-strong -fno-plt -Wconversion -O2 -g -DMIN_LOG_LEVEL=3 -Og -g -Wall -Wextra -pedantic -Wno-unused-parameter -Wstrict-prototypes -std=gnu99 -Wimplicit-fallthrough -Wvla -fstack-protector-strong -fdiagnostics-color=auto -Wno-array-bounds -DINCLUDE_GENERATED_DECLARATIONS -D_GNU_SOURCE -DNVIM_MSGPACK_HAS_FLOAT32 -DNVIM_UNIBI_HAS_VAR_FROM -I/home/h-michael/.cache/yay/neovim-git/src/neovim-git/build/config -I/home/h-michael/.cache/yay/neovim-git/src/neovim-git/src -I/usr/include -I/home/h-michael/.cache/yay/neovim-git/src/neovim-git/build/src/nvim/auto -I/home/h-michael/.cache/yay/neovim-git/src/neovim-git/build/include
Compiled by h-michael@x1

Features: +acl +iconv +jemalloc +tui
See ":help feature-compile"

   system vimrc file: "$VIM/sysinit.vim"
  fall-back for $VIM: "/usr/share/nvim"

Run :checkhealth for more info
```
When I run :UpdateRemotePlugins, it returns like this.
```
Encountered ModuleNotFoundError loading plugin at /home/h-michael/.cache/dein/repos/github.com/Shougo/deoplete.nvim/rplugin/python3/deoplete: No module named 'neovim'
Traceback (most recent call last):
  File "/home/h-michael/.local/lib/python3.7/site-packages/pynvim/plugin/host.py", line 135, in _load
    module = imp.load_module(name, file, pathname, descr)
  File "/usr/lib/python3.7/imp.py", line 244, in load_module
    return load_package(name, filename)
  File "/usr/lib/python3.7/imp.py", line 216, in load_package
    return _load(spec)
  File "<frozen importlib._bootstrap>", line 696, in _load
  File "<frozen importlib._bootstrap>", line 677, in _load_unlocked
ModuleNotFoundError: No module named 'neovim'
remote/host: python3 host registered plugins []
[...]

PR #877, opened 2018-11-18
Fix import pynvim if find pynvim module

h-michael:
Sorry #876 incomplete.


Shougo:
I have already fixed.

PR #879, opened 2018-11-22
Change `neovim` module to `pynvim`

e-kwsm:


rplugin/python3/deoplete/logger.py
```diff
@@ -48,16 +48,18 @@ def setup(vim, level, output_file=None):
         try:
             import pkg_resources
 
-            neovim_version = pkg_resources.get_distribution('neovim').version
+            pynvim_version = pkg_resources.get_distribution('pynvim').version
         except ImportError:
-            neovim_version = 'unknown'
+            pynvim_version = 'unknown'
+        except pkg_resources.DistributionNotFound:
```
e-kwsm: In my environment (Python 3.5.3 + setuptools 40.5.0), `pkg_resources.get_distribution` raises this exception if the argument is not found, literally.
I'm not sure if other exceptions are thrown on a different environment.
My question is: why not `except:`?

Shougo: I will do it.

Issue #887, opened 2018-12-05
buffer path is not working on second file open


ggsalas:
# Problems summary
`deoplete#file#enable_buffer_path` seems only work on some files (probably the firsrt one)

## Expected
complete path relative to current buffer 

## Environment Information

 * deoplete version (SHA1):

 * OS: macos Mojave 10.14.1 (18B75)
 * neovim/Vim `:version` output: v0.3.1

 * `:checkhealth` or `:CheckHealth` result(neovim only):
```

health#nvim#check
========================================================================
## Configuration
  - OK: no issues found

## Performance
  - OK: Build type: Release

## Remote Plugins
  - OK: Up to date

## terminal
  - INFO: key_backspace (kbs) terminfo entry: key_backspace=\177
  - INFO: key_dc (kdch1) terminfo entry: key_dc=\E[3~
  - INFO: $COLORTERM='truecolor'
```

## Provide a minimal init.vim/vimrc with less than 50 lines (Required!)
```
set runtimepath+=~/.config/nvim/plugged/deoplete.nvim/
let g:deoplete#enable_at_startup = 1
let g:deoplete#file#enable_buffer_path = 1
```
## How to reproduce the problem from neovim/Vim startup (Required!)
open a file: is working
open a second file: not work

## Screenshot (if possible)
![image](https://user-images.githubusercontent.com/8403993/49543883-21d3ce00-f8b8-11e8-9960-586622707523.png)

ggsalas:
Hi @Shougo Thanks for the fix. I have updated deoplete but I still having the issue. 

In the screenshot you can see that the completions are not inside the current file folder: 
![image](https://user-images.githubusercontent.com/8403993/50054730-ba254a80-0123-11e9-9879-a4dbe0a90a35.png)

Shougo:
Please see #893

ggsalas:
Thanks @Shougo! I have updated deoplete and is working perfect now.

PR #889, opened 2018-12-08
fixing file path completion when used with NERDTree

amunra81:
When used with NERDTree, filepath completion goes to the root of selected NERDTree directory instead of current buffer's filepath


Shougo:
What is your deoplete version?
I think it is the same with https://github.com/Shougo/deoplete.nvim/issues/887.
And it is already fixed.

amunra81:
Is the latest version, updated with Plug

Shougo:
Really?  It must be reproduce-able for me.

Please create the new issue for it with the minimal configuration and reproduce instructions.
If not, I will ignore your issue.

Shougo:
@amunra81 I have fixed the problem in the latest version.

PR #898, opened 2018-12-27
Removed print() messages

jshin:
Python's print() messages has now been displayed in `:messages` since ec014e3a8d53b7bbc8a3906013975157d09936a1 was committed.
If this is unintended behavior, please merge.


Shougo:
Thanks

Issue #914, opened 2019-01-12
The cursor jumps to the beginning fo the file when starting to edit a HTML property in file with multiple types


loomchild:
**Warning:  I will close the issue without the minimal init.vim and the reproduction instructions.**

# Problems summary

I have an issue when editing .vue files. The cursor jumps to the beginning fo the file when starting to edit a property in template/HTML part of the file.

It seems to be very similar problem to the one that was fixed before: https://github.com/Shougo/deoplete.nvim/issues/163. 

## Expected

The cursor should stay in the original position.

## Environment Information

 * deoplete version (SHA1): a2019e9f4f764a48a96a75e96b1b4f0bb79b4827 (master HEAD)

 * OS: Linux (Debian Linux 9.6, Linux 4.9.0)

 * neovim/Vim `:version` output:

```                                                                                                                                            
NVIM v0.4.0-62-g103e02191
Build type: RelWithDebInfo
LuaJIT 2.0.5
Compilation: /usr/bin/gcc-5 -Wshadow -Wconversion -U_FORTIFY_SOURCE -D_FORTIFY_SOURCE=1 -O2 -g -DMIN_LOG_LEVEL=3 -Og -g -Wall -Wextra -pedantic -Wno-unused-pa
rameter -Wstrict-prototypes -std=gnu99 -Wvla -fstack-protector-strong -fdiagnostics-color=auto -Wno-array-bounds -DINCLUDE_GENERATED_DECLARATIONS -D_GNU_SOURC
E -DNVIM_MSGPACK_HAS_FLOAT32 -DNVIM_UNIBI_HAS_VAR_FROM -I/home/travis/build/neovim/bot-ci/build/neovim/build/config -I/home/travis/build/neovim/bot-ci/build/n
eovim/src -I/home/travis/build/neovim/bot-ci/build/neovim/.deps/usr/include -I/usr/include -I/home/travis/build/neovim/bot-ci/build/neovim/build/src/nvim/auto
 -I/home/travis/build/neovim/bot-ci/build/neovim/build/include
Compiled by travis@travis-job-96499c6c-0595-42c9-80a3-53f4479b4eaf

Features: +acl +iconv +jemalloc +tui
See ":help feature-compile"

   system vimrc file: "$VIM/sysinit.vim"
  fall-back for $VIM: "/home/travis/build/neovim/bot-ci/build/neovim/build/nvim.AppDir/usr/share/nvim"

Run :checkhealth for more info```

[...]

PR #915, opened 2019-01-12
blacklist specific omnifunc also for CSS and JS

loomchild:


rplugin/python3/deoplete/source/omni.py
```diff
@@ -61,11 +61,13 @@ def _get_complete_position(self, context, current_ft, filetype):
                                            'Manual' and m is None):
                     continue
 
-                if filetype == current_ft and self._omnifunc in [
+                if ((filetype == current_ft or
+                        filetype in ['css', 'javascript'])
```
loomchild: Alternatively remove filetype condition (remove `((filetype == current_ft or filetype in ['css', 'javascript'])`) and rely only on blacklisted omnifunc's.

Shougo: I think specify omnifuncs names are better.

Shougo:
Why the feature is needed?   Please describe it.

Shougo:
Oh, OK.
It is for #914.

PR #924, opened 2019-02-02
Set equal flag

Shougo:
For https://github.com/neovim/neovim/pull/9566 test


Shougo:
It is merged in both neovim and Vim

PR #929, opened 2019-02-04
Filter for reordering kind types

reaysawa:
To be introduced by a custom option: `'kind_order_preference'`.

Attemps to reorder candidates from the source.

```
call deoplete#custom#option({
  \ 'kind_order_preference': {
  \   'javascript': ['!File', 'Function', 'Method'],
  \ }
})
```

- Remove any kind "File" - **"!" operator inverts filter**
- Order Function first
- Order Method second
- Order remaining kind types at the bottom

### Example:

`
call deoplete#custom#option({'kind_order_preference': {'javascript': ['!File', 'Function', 'Method']}})
`

#### Before
```
[
  {'word': 'Car', kind: 'File'},
  {'word': 'Door', kind: 'Method'},
  {'word': 'Open', kind: 'Function'},
  {'word': 'Car2', kind: 'File'}
]
```

#### After

```
[
  {'word': 'Open', kind: 'Function'},
  {'word': 'Door', kind: 'Method'}
]
```

## Missing
- [ ] Tests
- [ ] Documentation

@Shougo should the tests be done  like [test_matcher_fuzzy.py](https://github.com/Shougo/deoplete.nvim/blob/master/test/rplugin/python3/deoplete/test_matcher_fuzzy.py) (as in, does the API still work the same way)? Do I need to set custom options first with Vim?


Shougo:
I think it should be implemented by custom sorter.

The sorter name should be `sorter_kind`.

Shougo:
>  Documentation

The documentation is required.

rplugin/python3/deoplete/filter/converter_reorder_kind.py
```diff
@@ -0,0 +1,50 @@
...
+class Filter(Base):
+    def __init__(self, vim):
+        super().__init__(vim)
+
+        self.name = 'converter_reorder_kind'
+        self.description = 'Reorder candidates based on their kind'
+
+    def filter(self, context):
+        preferred_order = self.vim.call(
+            'deoplete#custom#_get_option', 'kind_order_preference'
```
Shougo: It should be `deoplete#custom#_get_filetype_option`.

autoload/deoplete/init.vim
```diff
@@ -213,6 +213,7 @@ function! deoplete#init#_option() abort
         \ 'ignore_sources': {},
         \ 'candidate_marks': [],
         \ 'max_list': 500,
+        \ 'kind_order_preference': {},
```
Shougo: The option name should be `kind_orders`.

Shougo:
And this feature is really useful?
Please describe it.

rplugin/python3/deoplete/filter/converter_reorder_kind.py
```diff
@@ -0,0 +1,50 @@
...
+            size = len(context['candidates'])
+            i = 0
+            while i < size:
+                if context['candidates'][i]['kind'] == kind:
+                    candidate = context['candidates'].pop(i)
+                    size -= 1
+                    if not disabled:
+                        new_candidates.append(candidate)
+                        new_candidates_len += 1
+                        # stop filtering if the maximum has been achieved
```
Shougo: This stop is really needed?

rplugin/python3/deoplete/filter/converter_reorder_kind.py
reaysawa: how do I specify options for filters?

rplugin/python3/deoplete/filter/converter_reorder_kind.py
reaysawa: Not really needed but there is no point to go forward if the rest will not be seen, given that the algorithm is already O(n) (for the number of candidates * the number of filters) + the cost of popping from the array.

That being said, now I notice it might modify the `context['candidates']` array (thus removing other options that might be useful?), so I'll make a copy and manipulate it instead.

reaysawa:
@Shougo 
> And this feature is really useful?
> Please describe it.

I find it useful because LSP will often try to "guess" autocompletion, but sometimes the suggestions are useless. The kind "File" is just random words from the file and I can use 'buffer' or 'around' sources for that.

For instance, here I am ordering **1.Method and 2.Field first so they appear at the top** (File should be filtered out, but the code is bugged so File doesn't get completely removed).

image

If I don't reorder them, I get a bunch of garbage at the top and when filtering too.

image

---

deoplete-go does this with

https://github.com/deoplete-plugins/deoplete-go#gdeopletesourcesgosort_class

`let g:deoplete#sources#go#sort_class = ['package', 'func', 'type', 'var', 'const']`

But not all sources have that feature.

reaysawa:
Also, currently it's both a filter (filters out '!name' kinds) and a sorter. Would it be better if more candidate fields could be filtered, or is just "kind" good enough?

```
call deoplete#custom#option({
  \ 'candidate_attribute_filters': {
  \   'javascript': { 'kind': ['!File', 'Function', 'Method'] }
  \ }
})
```

Shougo:
I think to add `deoplete#custom#filter()` is better instead to add new option.

> Would it be better if more candidate fields could be filtered, or is just "kind" good enough?

I think more field could be filtered.
So the filter should be named `converter_reorder_attr`.

> I have a good idea for tests and documentation if everything is fine now.

Please add them.

reaysawa:
Changes ready @Shougo with tests passing : ^ )

Shougo:
I need to add `denite#custom#filter()`.

Please wait.

Shougo:
I have added `deoplete#custom#source()`.  Please use it.

doc/deoplete.txt
```diff
@@ -154,6 +154,93 @@ ignore_case
 ignore_sources
 		It is a dictionary to decide ignore source names.
 		The key is filetype and the value is source names list.
+		Default value: {}
+
+						*deoplete-options-attrs_order*
+attrs_order
```
Shougo: `attrs_order` should be `converter_reorder_attr` specific var like `delimiters` in `converter_auto_delimiter`.

rplugin/python3/deoplete/filter/converter_reorder_attr.py
```diff
@@ -0,0 +1,69 @@
...
+                        size -= 1
+                        if not disabled:
+                            new_candidates.append(candidate)
+                            new_candidates_len += 1
+                            # stop filtering if the maximum has been achieved
+                            if new_candidates_len == max_list_size:
+                                return new_candidates
+                    i += 1
+
+            # add remaining at the bottom
```
Shougo: `# Add remaining`

rplugin/python3/deoplete/filter/converter_reorder_attr.py
```diff
@@ -0,0 +1,69 @@
...
+            new_candidates.extend(context_candidates)
+            # go to the next attribute with the new list order
```
Shougo: `# Go to the next`

rplugin/python3/deoplete/filter/converter_reorder_attr.py
```diff
@@ -0,0 +1,69 @@
...
+            context_candidates = new_candidates
+
+        return new_candidates
+
+    def filter(self, context):
+        preferred_order_attrs = self.vim.call(
+            'deoplete#custom#_get_option', 'attrs_order'
+        ).get(context['filetype'])
```
Shougo: It should `get(context['filetype'], [])`

Shougo:
Please add the documentation for `converter_reorder_attr`.

reaysawa:
@Shougo sorry for the delay. I have updated the code to use `deoplete#custom#filter` (and also the documentation).

I'll be looking more actively to be able to merge the PR as soon as possible.

Shougo:
```
Filter custom variables:
converter_attrs
```

is needed.

doc/deoplete.txt
```diff
@@ -783,6 +784,93 @@ converter_auto_delimiter
...
+		The example above says: For Ruby files, I want to reorder the
+		candidates:
+		1. Candidates which have a 'kind' attribute containing 'Class'
+		will be filtered out of the list.
+		After they have been removed, proceed to the next attribute.
+		2. Candidates which have a 'abbr' attribute containing
+		'ExtensionClass' will be moved to the top of the list.
+		3. Any other candidates appear at the bottom, if there is
+		space.
+
```
Shougo:
```
Filter custom variables:
convertert_attrs
```

reaysawa: @Shougo Hi, I'm not sure what that means. Do you mean I should make the customisation through a variable, instead of passing the filetypes dictionary directly?

```
deoplete#custom#filter(converter, converter_attrs, { attrs })
```


Like this? If I haven't understood it and you find it easier, feel free to push your own fixes to this branch!

Shougo: OK.

PR #958, opened 2019-04-02
Improve events performance

Shougo:


takkii:
env, Windows10, kaoriya-vim8 GitHub latest.

I checked the operation in the callback branch.
There was nothing wrong with the :message. It does not seem to be damaged.

LGTM!

Shougo:
Thank you for the review!

PR #807, opened 2018-07-14
WIP: add failing test for #800, nvim test fixture

blueyed:

PR #983, opened 2019-05-29
ci: Travis: fix PYTEST_ADDOPTS

blueyed:
The leading "-" turned this into the "-r" option, which happily consumed
all the following chars as reportchars then.


blueyed:
Heh, was/is already included in https://github.com/Shougo/deoplete.nvim/pull/807/commits/3bd03a0975fdca4cbb21efd631291e95462bbaad - but good to merge it here already.

PR #997, opened 2019-08-01
Suppress key error.

Milly:
Key error is raised at completion, if unavailable source name set to `deoplete#custom#buffer_var`.

Ex.
```vim
call deoplete#custom#buffer_var('foo', 'bar', v:true)
```


Shougo:
Thank you for the reporting.
But my fix is better.

PR #1020, opened 2019-09-21
Prepare to support textEdit

hrsh7th:
Hi, Thanks for provided deoplete.nvim.
I like deoplete because it's fast.

This PR aims to prepare to avoid conflicts to `user_data` from other plugin.

For example, When support completion-item's `textEdit` that defined in LSP spec, it is difficult maybe.

The reason is that deoplete cause error if `user_data` is not string.
(`vim-lsp` assign dictionary to `user_data`)

I don't know the reason why deoplete checks `user_data`.
I could not find the assignment to `user_data` by deoplete itself.

### DIFF
--- a/autoload/deoplete/handler.vim
+++ b/autoload/deoplete/handler.vim
@@ -297,7 +297,12 @@ function! s:on_complete_done() abort
   call deoplete#handler#_skip_next_completion()
 
   if get(v:completed_item, 'user_data', '') !=# ''
-    call s:substitute_suffix(json_decode(v:completed_item.user_data))
+    try
+      if type(v:completed_item.user_data) == type('')
+        call s:substitute_suffix(json_decode(v:completed_item.user_data))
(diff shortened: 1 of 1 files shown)


Shougo:
> I don't know the reason why deoplete checks user_data.
I could not find the assignment to user_data by deoplete itself.

Because, LanguageClient-neovim uses `user_data`.

rafi:
Cool, this will make other completefunc plugins to work with deoplete, for example [kizza/actionmenu.nvim](https://github.com/kizza/actionmenu.nvim).

@Shougo Maybe deoplete can have a signature in `user_data` to know if it came from deoplete or a 3rd-party plugin?

Shougo:
> @Shougo Maybe deoplete can have a signature in user_data to know if it came from deoplete or a 3rd-party plugin?

It is impossible.

PR #1036, opened 2019-11-11
Fix docs regarding default converters

ertomentis:
Fixed documentation regarding default converters, as specified in rplugin/python3/deoplete/base/source.py


Shougo:
Thanks

PR #1041, opened 2019-11-20
python version check logic was inverted. probably typo from denite code

bazilek:
seems like typo from denite code. logic inside #init#_python_version_check() differs


Shougo:
I have fixed it.

Shougo:
Thanks.  But my fix is better...

PR #1047, opened 2019-12-14
Remove `encoding` and `unicode_errors` option from Packer and Unpacker

zchee:
Remove the `encoding` and `unicode_errors` option from `msgpack.Packer`,  and remove the `encoding` option from `msgpack.Unpacker`.
ref:
- msgpack/msgpack-python@e1ed004
- msgpack/msgpack-python@e419cd8

But both commits were 11 days ago and very recent commits (I've always `pip install git+http://...` so this problem occurred)

I don't know that can keep backward compatibility without that option, but this problem undoubtedly occurs when a new release of msgpack-python.
So, merge timing leaves you.

### DIFF
--- a/rplugin/python3/deoplete/child.py
+++ b/rplugin/python3/deoplete/child.py
@@ -40,12 +40,9 @@ class Child(logger.LoggingMixin):
         self._source_errors: typing.Dict[str, int] = defaultdict(int)
         self._prev_results: typing.Dict[str, Result] = {}
         self._unpacker = msgpack.Unpacker(
-            encoding='utf-8',
             unicode_errors='surrogateescape')
         self._packer = msgpack.Packer(
-            use_bin_type=True,
-            encoding='utf-8',
-            unicode_errors='surrogateescape')
+            use_bin_type=True)
         self._ignore_sources: typing.List[typing.Any] = []
 
     def main_loop(self, stdout: typing.Any) -> None:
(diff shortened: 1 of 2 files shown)


Shougo:
If the new version of msgpack-python is released, it should be merged.
Please wait until it.

zchee:
@Shougo I see, thanks quickly.
Nevermind, I'll use this PR with `dein#local` until has been merged this PR :)

liammccartney:
Hi @Shougo & @zchee!

I was setting up deoplete today and got tripped up by the fact that msgpack-python was installed at the latest version, [v1.0.0](https://github.com/msgpack/msgpack-python/releases/tag/v1.0.0). 
This release includes the removal of the `encoding` kwarg ([breaking changes listed here](https://github.com/msgpack/msgpack-python#major-breaking-changes-in-msgpack-10)).

Fortunately I found this issue and I was able to work around it! Anyway, I wanted to chime in b/c I think it might be time to consider merging this change, or something else to handle msgpacker's breaking changes  :) 

Thank you!

Shougo:
Hm... I have updated but it does not work...

Shougo:
msgpack-python 1.0.0 is `*not*` released in pip.
And it breaks in old msgpack-python environments.

So I need to revert the changes.

Shougo:
Hm..  I need to install `msgpack` package by pip...

Shougo:
I need to add msgpack version check function.
Please wait.

Shougo:
Please use #1070 instead.
I need to test it for a while.

liammccartney:
@Shougo shoot! I'm sorry if I mislead you. 
I'm confused as to how I wound up with v1.0.0 of msgpack-python

PR #1061, opened 2020-01-29
Fix not to complete "-1" for every auto-completion

ujihisa:
* https://gyazo.com/059c77dfa276072776d9b7aee1eafdfa
* Looks like this fix caused this bug https://github.com/Shougo/deoplete.nvim/pull/1058


Shougo:
scriptencoding is needed.  I have fixed the error.

PR #1066, opened 2020-02-15
feat: add manual_menu option

### LINKED ISSUE #1065: Please provide a way to use it without popup menu
I hate to hide adjacent lines of code by completion popup menu, this is just not an acceptable workflow to me. At same time I like how this plugin works at all other aspects. I've spent two days trying a lot of different configurations based on `deoplete#manual_complete()` recipe in the doc but failed, and looks like I'll have to give up and try some other plugin for completion. :disappointed: 

I suppose this plugin may be somehow improved to enable workflow I need, please consider such a feature:
* Completion still gather data async but no menu is visible on screen and no completion is applied automatically just because I've typed some prefix string.
* On pressing `<Tab>` completion menu may or may no be shown - both cases are acceptable, but if it's shown then it must disappears right after some completion will be selected or all completions will be rejected (like by typing non-keyword symbol) and must not be shown again automatically (without pressing `<Tab>`).
* On pressing `<Tab>` (same key press as mentioned in previous item, not second one) first completion item must be used, i.e. it should work just like pressing `<C-n>` when menu is shown. Pressing `<Tab>` again and again should work like `<C-n>`.

I belive troubles I had was related to #1006 - I've used suggestions from LSP (provided by vim-go), and while at a glance it works next when I've tried to `<BS>` several times after accepting initial successful completion and press `<Tab>` once again I got either no result, or sometimes valid result intermixed with invalid (wrong prefix string with head matcher). Sometimes it works okay, but most often it doesn't. Yeah, I suppose this is probably some sort of async/timing related issue, and probably we need to implement optional sync mode for deoplete to make it work reliably with mentioned above workflow.

powerman:
Closes #1065 

Example `vimrc` setup with extra feature to auto-insert first completion:
```vim
call deoplete#custom#option('manual_menu', v:true)

inoremap <silent><expr> <TAB>
    \ pumvisible() ? "\<C-n>" :
    \ <SID>check_back_space() ? "\<Tab>" :
    \ <SID>manual_complete()
function! s:check_back_space() abort
    let col = col('.') - 1
    return !col || getline('.')[col - 1]  =~# '\s'
endfunction
function! s:manual_complete()
    call timer_start(100, {-> feedkeys("\<C-n>", 'in')})
    return deoplete#manual_complete()
endfunction
```

### DIFF
--- a/autoload/deoplete.vim
+++ b/autoload/deoplete.vim
@@ -57,6 +57,7 @@ function! deoplete#manual_complete(...) abort
   call deoplete#init#_prev_completion()
 
   " Start complete.
+  let g:deoplete#_manual = v:true
   return "\<C-r>=deoplete#mapping#_rpcrequest_wrapper("
         \ . string(get(a:000, 0, [])) . ")\<CR>"
 endfunction
--- a/autoload/deoplete/handler.vim
+++ b/autoload/deoplete/handler.vim
@@ -82,6 +82,7 @@ function! deoplete#handler#_check_omnifunc(context) abort
         \ || index(blacklist, &l:omnifunc) >= 0
         \ || prev.input ==# a:context.input
         \ || s:check_input_method()
+        \ || deoplete#custom#_get_option('manual_menu') && !g:deoplete#_manual
     return
   endif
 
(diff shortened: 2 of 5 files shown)


Shougo:
Thank you for the PR.
I have checked your implementation.

But my PR is better.

https://github.com/Shougo/deoplete.nvim/pull/1067

Please test it instead.

PR #1067, opened 2020-02-15
Add auto_complete_popup option

### LINKED ISSUE #1065: Please provide a way to use it without popup menu

Shougo:
Fix #1065

### DIFF
--- a/autoload/deoplete.vim
+++ b/autoload/deoplete.vim
@@ -46,6 +46,9 @@ function! deoplete#send_event(event, ...) abort
         \ {'event': a:event, 'sources': sources})
 endfunction
 
+function! deoplete#complete() abort
+  return deoplete#mapping#_dummy('deoplete#mapping#_complete')
+endfunction
 function! deoplete#auto_complete(...) abort
   return deoplete#handler#_completion_begin(get(a:000, 0, 'Async'))
 endfunction
--- a/autoload/deoplete/handler.vim
+++ b/autoload/deoplete/handler.vim
@@ -65,13 +65,18 @@ function! deoplete#handler#_do_complete() abort
   let prev.complete_position = context.complete_position
   let prev.linenr = line('.')
 
+  let auto_popup = deoplete#custom#_get_option(
+        \ 'auto_complete_popup') !=# 'manual'
+
   if context.event ==# 'Manual'
     let context.event = ''
-  elseif !exists('g:deoplete#_saved_completeopt')
+  elseif !exists('g:deoplete#_saved_completeopt') && auto_popup
     call deoplete#mapping#_set_completeopt()
   endif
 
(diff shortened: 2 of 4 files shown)


zchee:
# [Codecov](https://codecov.io/gh/Shougo/deoplete.nvim/pull/1067?src=pr&el=h1) Report
> Merging [#1067](https://codecov.io/gh/Shougo/deoplete.nvim/pull/1067?src=pr&el=desc) into [master](https://codecov.io/gh/Shougo/deoplete.nvim/commit/6e01000280edc9c9c56ef7b1d59977fc93e3bf4c&el=desc) will **not change** coverage by `%`.
> The diff coverage is `n/a`.

```diff
@@           Coverage Diff           @@
##           master    #1067   +/-   ##
=======================================
  Coverage   49.43%   49.43%           
=======================================
  Files          18       18           
  Lines         969      969           
=======================================
  Hits          479      479           
  Misses        490      490           
```

zchee:
oh? why I commented? lol

Shougo:
@powerman no response?

Shougo:
OK.  So I will merge it later.

powerman:
@Shougo Sorry, I didn't received emails from GitHub with your comments - no idea why, usually I get all emails from github. Anyway, at a glance looks like your change works for me, thanks a lot!

PR #1069, opened 2020-02-18
Revert "Revert "Remove `encoding` and `unicode_errors` option from Packer and Unpacker""

Shougo:
Reverts Shougo/deoplete.nvim#1068


Shougo:
Fixed by #1070

PR #1101, opened 2020-06-06
Typing parent

Milly:
Fix typing in *parent.py*.
<del>The implementation has not changed.</del>

Change platform detection code, because MyPy recognized [PEP0484](https://www.python.org/dev/peps/pep-0484/#version-and-platform-checking).
See: https://github.com/python/mypy/issues/8547


Shougo:
Thanks!

PR #1104, opened 2020-06-16
[RDY] Fix is_async confirmation

### LINKED ISSUE #1013: deoplete now automatically completes without permission
Seems to be exactly as in #440, but the problem is back.

Note this only seems to be in `master`, it seems fine in 5.1.  Unfortunately I am still using Vundle which always pulls master.

Shougo:
Fix #1013


Shougo:
Merged.

PR #1125, opened 2020-09-02
README has inverted logic for vim-plug

isovector:
The README asks for nvim packages under vim-plug when nvim isn't running; or at least, this logic is inverted compared to the next block.


Shougo:
The inverted fix is wrong.
You should read the code carefully.

PR #1146, opened 2020-11-10
[RFC] Add recently_used sort feature

Shougo:


Shougo:
More testers are needed.

PR #1163, opened 2021-02-15
added msgpack requirement in README

CosmosAtlas:
When enabling the plugin via the current readme on GitHub, completion for files do not work.

After reading the docs in vim through `help deoplete` a new package `msgpack/msgpack-python` was required. Thus adding this information in the README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,6 +47,7 @@ https://github.com/Shougo/deoplete.nvim/releases/tag/5.2
 For vim-plug
 
 ```viml
+Plug 'msgpack/msgpack-python'
 if has('nvim')
   Plug 'Shougo/deoplete.nvim', { 'do': ':UpdateRemotePlugins' }
 else


Shougo:
You need to read this.

```
### Requirements

deoplete requires Neovim or Vim8 with `if_python3`.

If `:echo has("python3")` returns `1`, then you have python 3 support;
otherwise, see below.

You can enable Python3 interface with pip:

    pip3 install --user pynvim

Please install nvim-yarp and vim-hug-neovim-rpc for Vim8.

- <https://github.com/roxma/nvim-yarp>
- <https://github.com/roxma/vim-hug-neovim-rpc>
```

`msgpack` is installed as `pynvim` requirements.  So I don't add it on README.

CosmosAtlas:
Ok, I guess I'll add more context.

The file completion was working until recently (~2 weeks). I'm certain that my vim (and neovim) have python support. Since I really need file completion, so I read the docs to see if I was missing anything. And after explicitely having  `msgpack/msgpack-python` file completion is working correctly again. 

So maybe a potential bug elsewhere. Anyway, thanks for the input.

Shougo:
I think deoplete upgrade is needed.
I have fixed several file completion problems recently.

PR #1176, opened 2021-05-16
Fix `input_patterns` source option not being recognised

habibalamin:
`input_pattern` is recognised, so I think `input_patterns` should be as well. The inability to set this source option from Vim config is really confusing when `input_pattern` can be set.

In particular, I needed to set a slightly different `input_pattern` for Haskell when using LanguageClient-neovim.


Shougo:
Thanks.

Issue #1152, opened 2020-12-15
Doesn't work with vim package manager.


smekkley:
How to reproduce:
1. cd ~/.vim/pack/git-plugins/start
2. git clone https://github.com/Shougo/deoplete.nvim.git
3. Start vim and you get exceptions and no completion works.

Shougo:
Please don't ignore the issue template.
Please read other issues and compare with yours.

Shougo:
> Start vim and you get exceptions and no completion works.

What is the exceptions?
And why don't you upload the minimal vimrc?

smekkley:
vimrc is empty. I might paste the exception if I install it again. It's not functional for built-in vim package manager users in any case.

Shougo:
deoplete doesn't work if vimrc is empty.
Why don't you read the documentation?

smekkley:
Oh, it's supposed to crash just by git checkout? I might do that when I try this plugin next time.

Shougo:
I don't know.
I don't work for it without minimal vimrc.

Issue #990, opened 2019-07-04
Installing deoplete on Vim


YuhkiYano:
Can we install deoplete on Vim not using package manager plugins such as NeoBundle, Bundle, dein and so on? I would like to install it without the use of package manager as much as possible.

Shougo:
It is possible.

For example:

```vim
set rtp+=~/work/deoplete.nvim/
set rtp+=~/work/nvim-yarp
set rtp+=~/work/vim-hug-neovim-rpc/
let g:deoplete#enable_at_startup = 1
```

PR #1177, opened 2021-05-17
Installation using built-in package manager

klimeryk:
Vim 8 and NeoVim come with a [built-in package manager](https://vimhelp.org/repeat.txt.html#packages). However, the way it works is that it loads all the "auto-start" plugins _after_ loading the configuration file. So, Deoplete cannot be configured using the `deoplete#custom#option` function, as it's not available yet.

The solution is to make Deoplete an optional package and load it explicitly inside the configuration file.

This should prevent confusion like in https://github.com/Shougo/deoplete.nvim/issues/1152. Related issue: https://github.com/Shougo/deoplete.nvim/issues/990 (when not using _any_ package managers).

Thanks for your work on this great plugin!

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,7 +44,7 @@ instead.
 
 https://github.com/Shougo/deoplete.nvim/releases/tag/5.2
 
-For vim-plug
+### vim-plug
 
 ```viml
 if has('nvim')
@@ -57,7 +57,7 @@ endif
 let g:deoplete#enable_at_startup = 1
 ```
 
-For dein.vim
+### dein.vim
 
 ```viml
 call dein#add('Shougo/deoplete.nvim')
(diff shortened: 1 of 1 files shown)


Shougo:
Why do you write the instruction in README only??
Please add it in the help.

README.md
```diff
@@ -68,7 +68,53 @@ endif
 let g:deoplete#enable_at_startup = 1
 ```
 
-For manual installation(not recommended)
+### Built-in Vim 8/NeoVim package manager (pack feature)
+
+Becasue Deoplete uses functions for configuration, we have to ensure it is loaded. So we put it as an *optional* package (`opt` subdirectory of your pack) and load it explicitly in the configuration file:
```
Shougo, line 73: Because

README.md
```diff
@@ -68,7 +68,53 @@ endif
...
+
+#### Vim 8
+```
```
Shougo, line 76: Please add a blank line.

README.md
```diff
@@ -68,7 +68,53 @@ endif
...
+let g:deoplete#enable_at_startup = 1
+
+" This actually loads the plugin
+packadd deoplete.nvim
+
+" So here it's safe to call things like call deoplete#custom#option
+call deoplete#custom#option('ignore_sources', {'_': ['around', 'buffer']})
+```
+
+#### Generate the remote plugin manifest
```
Shougo, line 108: It is neovim only instruction.  It should be described.

README.md
```diff
@@ -68,7 +68,53 @@ endif
 let g:deoplete#enable_at_startup = 1
 ```
 
-For manual installation(not recommended)
+### Built-in Vim 8/NeoVim package manager (pack feature)
```
Shougo, line 71:
```
### Built-in Vim 8/NeoVim package manager(pack feature)
```
?

Because

```
### Manual installation(not recommended)
```

is later.

Shougo:
@klimeryk Ping.  No response?  If so, I will close the issue.

klimeryk:
Sorry, for the late response! 🙇 I did not think about adding the installation instructions to the help file since that one is available only after the plugin is installed?

Either way, I've since stopped using deoplete as it seemed to have had a significant impact on the battery life. I'll close this PR - at least if someone finds the linked issues or searches the issue/PR history, they should find some hints on how to proceed for the built-in package manager.

Thanks for having a look and thanks for all your work on deoplete! 🙇

Shougo:
>Either way, I've since stopped using deoplete as it seemed to have had a significant impact on the battery life.

Oh, OK.

klimeryk:
> Either way, I've since stopped using deoplete as it seemed to have had a significant impact on the battery life.

Just to be clear - I did not dig deeper what was the exact issue. I just noticed that my new MacBook Pro had ~3 hours battery life when working with just neovim from the terminal. Which seemed very odd. I figured it's one of the plugins causing it - and narrowed it down to deoplete. But I did not dig deeper - maybe it was the integration with ale (I still have `ale` itself running, though, without a significant battery drain). Afterwards, I had easily ~8 hours of battery time.

I just wanted to clarify this - once I realized it's deoplete, it actually made me realize I don't use autocompletion results anyway, so I just removed the plugin without digging deeper 🙇

Shougo:
I think ale deoplete source has battery drain.

https://github.com/dense-analysis/ale/blob/master/rplugin/python3/deoplete/sources/ale.py

You can disable ale source by `ignore_source` option.

```vim
call deoplete#custom#option('ignore_sources', {'_': ['ale']})
```

PR #1187, opened 2021-07-26
feat: Add the post update hook to dein readme

TeoDev1611:
Add the post update hook for improve the experience in dein :p


Shougo:
It is not needed.  Because dein executes it automatically.

PR #1198, opened 2022-08-15
Add missing negation in vim version validation

nospam2678:
Commit 610daca4 intended to elevate the vim version check. Unfortunately
the logic was reversed, effectively ensuring the opposite of what is
desired.


Shougo:
Thanks.

PR #1199, opened 2022-08-21
Suppress error message when no pattern matches

larskotthoff:
Commit f7b8364213e1c33164cbf97c5e1c7c5913bf0a97 changed the way the completion function is called, resulting in a "Pattern not found" and "Press enter or type command to continue" message when there were no matches. Calling the command silently fixes this.

### DIFF
--- a/autoload/deoplete/mapping.vim
+++ b/autoload/deoplete/mapping.vim
@@ -18,7 +18,7 @@ function! deoplete#mapping#_init() abort
   noremap  <silent> <Plug>+  <Nop>
 endfunction
 function! deoplete#mapping#_dummy(func) abort
-  return "\<Cmd>call ".a:func."()\<CR>"
+  return "\<Cmd>silent call ".a:func."()\<CR>"
 endfunction
 function! s:check_completion_info(candidates) abort
   if !exists('*complete_info')


Shougo:
Hi.  I have tested in the latest version.  But I cannot reproduce your messages.

> Commit https://github.com/Shougo/deoplete.nvim/commit/f7b8364213e1c33164cbf97c5e1c7c5913bf0a97 changed the way the completion function is called, resulting in a "Pattern not found" and "Press enter or type command to continue" message when there were no matches. Calling the command silently fixes this.

Please upload the minimal init.vim.  I will test it later.

Shougo:
I don't merge it until the problem is reproduced for me.

larskotthoff:
Thanks for checking -- for some weird reason, it looks like this behavior is triggered by

    set noshowmode

I'm not sure why, but this is probably sufficiently niche that it doesn't need to be addressed. Feel free to close.

Shougo:
OK.  It is reproduced.

sirfz:
FYI, I'm still getting error (in Python) on each key stroke in the latest master (`Error detected while processing function deoplete#mapping#_prev_complete[13]`). It renders vim unusable, reverted to commit `33ed4fa0cd704999f9f74b37640bf6d7334bac37` and it's back to normal.

Shougo:
It seems your configuration error.
Because deoplete does *not* define the `pclose` autocmd.
Please create the new issue and upload the minimal init.vim.  If not, I just ignore your report.

Shougo:
> Update: the revert to that commit didn't fix the problem

I know. Because it is just your configuration error.

sirfz:
I haven't touched my configuration in years so it might be anything for all I know. Will try to figure it out when I have time and open an issue if it's relevant to deoplete.

PR #1205, opened 2023-07-14
Update path for cpsm_py module

phongnh:
I tried to use `matcher_cpsm` in deoplete.nvim, I found that https://github.com/nixprime/cpsm now install `cpsm_py.so` in `autoload` folder instead of `bin`.

Please help to review this PR, and merge it if it is helpful, thanks.


Shougo:
Thanks.

